PR #144, opened 2015-03-11
'lib/log: save dlpi_names and iterate over them after dl_iterate_phdr()

dpejesh:
Some platforms (eg. FreeBSD 10+) don't support calling dlopen() from
within a dl_iterate_phdr() callback; so save all of the dlpi_names to
a list and iterate over them afterwards.

This fixes a deadlock which occurs in dlopen() when called from
within dl_iterate_phdr() on FreeBSD 10+ when linked to libthr.


chrissie-c:
Merged. Thanks.

PR #161, opened 2015-07-20
Fix: valgrind "invalid file descriptor" warning

kgaillot:
This migrates the libtool soname versioning to use -version-info and throws in some minor other tweaks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,10 @@
 language: c
 compiler:
   - gcc
-before_install: sudo apt-get install check splint
-install:
- # Deal with issue on Travis builders
- # https://github.com/travis-ci/travis-cookbooks/issues/155
- - "sudo rm -rf /dev/shm && sudo ln -s /run/shm /dev/shm"
+addons:
+        apt:
+                packages:
+                  - check
+                  - splint
 script: ./autogen.sh && ./configure && make check && make distcheck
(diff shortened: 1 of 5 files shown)


kgaillot:
I decided to give a shot at migrating libtool to -version-info, and migrating travis CI from their "legacy infrastructure" to the supposedly faster stuff described at http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade

Packages built with this installed on my RHEL 7 cluster as /usr/lib64/libqb.so.0.17.2 with no complaints, and the travis build passed with "This job is running on container-based infrastructure". :)

I'm currently running 100 random CTS tests with nodes configured with various values of PCMK_debug, PCMK_trace_functions, PCMK_trace_files, and PCMK_trace_formats. So far so good.

I pushed a v0.17.2 tag with this request. I don't know whether merging pull requests pulls in tags, so if we really want to release this as v0.17.2, make sure it's tagged.

kgaillot:
Rebased, and updated fd test.

jnpkrn:
Thanks for 780f893, otherwise I would be covering that :)

kgaillot:
Yep that was an annoyance in valgrind reports :)

PR #163, opened 2015-08-28
Add online documentation for v0.17.1

kgaillot:


kgaillot:
I updated the helper script for building online documentation (which will be in a separate pull request for the master branch). Since by happenstance the documentation for 0.17.1 was never generated, I tested it using that, and this is the result.

Once we clear up the v0.17.2 tagging issue, we can run the new script for 0.17.2.

PR #172, opened 2016-01-21
ipc: update file permissions on created sockets

dpejesh:
Follow-up to my last pull request, the mode should have been 666 instead of 660.


chrissie-c:
Hi. Sorry, I won't merge a patch that forces the sockets to be publicly writeable. Processes should at least be in the same group.

PR #175, opened 2016-02-03
Format annotations

jnpkrn:


chrissie-c:
Merged, thanks. I'm going to close issue #143 even though it's not a full solution as I think that might be a major project!

PR #176, opened 2016-02-12
Changing syslog identifier + testing syslog sink

jnpkrn:


jnpkrn:
Do not pull yet, please.  I think there can be a more versatile interface with `qb_log_ctl2`.

chrissie-c:
Closed in preference to #177

PR #187, opened 2016-02-24
Patches from Svante Signell

jnpkrn:
See:
- [1st @ ClusterLabs](http://oss.clusterlabs.org/pipermail/developers/2016-February/000154.html)
  or [1st @ gmane](http://article.gmane.org/gmane.comp.clustering.clusterlabs.devel/115)
- [2nd @ ClusterLabs](http://oss.clusterlabs.org/pipermail/developers/2016-February/000155.html)
  or [2nd @ gmane](http://article.gmane.org/gmane.comp.clustering.clusterlabs.devel/116)

Ensured that committer = author as I have nothing to do with this per se.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -335,6 +335,11 @@ case "$host_os" in
 		CP=rsync
 		AC_MSG_RESULT([Solaris])
 	;;
+	*gnu*)
+		AC_DEFINE_UNQUOTED([QB_GNU], [1],
+				   [Compiling for GNU/Hurd platform])
+		AC_MSG_RESULT([GNU])
+	;;
 	*)
 		AC_MSG_ERROR([Unsupported OS? hmmmm])
 	;;
--- a/lib/ipc_setup.c
+++ b/lib/ipc_setup.c
@@ -286,7 +286,7 @@ qb_ipcc_stream_sock_connect(const char *socket_name, int32_t * sock_pt)
 #endif
 
 #if defined(QB_LINUX) || defined(QB_CYGWIN) || defined(QB_GNU)
-	snprintf(address.sun_path + 1, UNIX_PATH_MAX - 1, "%s", socket_name);
+	snprintf(address.sun_path, sizeof(address.sun_path), "%s", socket_name);
 #else
 	snprintf(address.sun_path, sizeof(address.sun_path), "%s/%s", SOCKETDIR,
 		 socket_name);
@@ -535,7 +535,7 @@ qb_ipcs_us_publish(struct qb_ipcs_service * s)
 
 	qb_util_log(LOG_INFO, "server name: %s", s->name);
 #if defined(QB_LINUX) || defined(QB_CYGWIN) || defined(QB_GNU)
-	snprintf(un_addr.sun_path + 1, UNIX_PATH_MAX - 1, "%s", s->name);
+	snprintf(un_addr.sun_path, sizeof(un_addr.sun_path), "%s", s->name);
 #else
 	{
 		struct stat stat_out;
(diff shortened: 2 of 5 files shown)


kgaillot:
Replied to the list thread as well with this:

* The leading null byte in sun_path (the "+ 1" when using it) indicates an abstract socket (see unix(7)). If Hurd doesn't support those, it'll need a separate block -- we shouldn't change the behavior for Linux/Cygwin.

* I don't think the unlink() will make sense with abstract sockets.

* Not necessary with this request, but it would be nice if we could have ./configure define semantic macros like HAVE_ABSTRACT_SOCKET or something. Or at least #define them ourselves so we don't have to repeat "defined(...) || ..." all the time.

gnu-srs:
Sorry for commenting here as well as replying to the mailing list. I don't know where to ad a comment:
1) Here -> automatically to the ML?
2) To the ML-> Here?
3) Neither of the above.

 I added this comment to the pull request:
> 
> * The leading null byte in sun_path (the "+ 1" when using it) indicates
> an abstract socket (see unix(7)). If Hurd doesn't support those, it'll
> need a separate block -- we shouldn't change the behavior for Linux/Cygwin.

Now I see why you use the + 1 in sun_path, thanks. However, the second patch was
derived under GNU/Linux, where the sun_path was found to be NULL, as by the
comment. It FTBFS without the unlink statement with the second patch. The
unpatched code in Debian version 0.17.2.real-4 FTBFS with gcc-4.9.1-16, as
written in Debian bug #803766. How is the abstract socket selected in favor of
the regular socket in GNU/Linux?

> * I don't think the unlink() will make sense with abstract sockets.

Same here, the unlink was is needed with the latest gcc: 5.3.1-8 :( Otherwise
ipc.test fails.

> * Not necessary with this request, but it would be nice if we could have
> ./configure define semantic macros like HAVE_ABSTRACT_SOCKET or
> something. Or at least #define them ourselves so we don't have to repeat
> "defined(...) || ..." all the time.

That would definitely be needed, especially if abstract sockets are not used in
GNU/Linux, as I probably experienced, see above.

kgaillot:
There's no automatic link between here and the list; I just copied-and-pasted my comment in case you weren't following this request. We can just use this request now.

The leading null byte before the path is how Linux decides that it should be an abstract socket. Abstract sockets don't have any presence on the filesystem, so the path is just a label in that case.

That's interesting about the newer gcc. I wonder why we haven't seen that in Fedora builds. If you revert the second commit, what error does the bind() return?

gnu-srs:
Without the Linux patch:
check_ipc.c:1464:F:ipc_max_dgram_size:test_ipc_max_dgram_size:0: Assertion 'init == try' failed: init == 331264, try == 425728

Need some time to find out the return code of bind().

gnu-srs:
gcc-5.real (Debian 5.3.1-8)

gnu-srs:
Just a note: From man(7) unix:  SCM_CREDENTIALS and the abstract namespace were introduced  with  Linux  2.2  and  should  not  be used in portable programs.  Any specific reason you want to use abstract sockets for Linux/Cygwin?

kgaillot:
That was before my time. Per git history, it was done as part of the switch to datagram sockets "to prevent partial sends and gettting stuck in retry loops". Abstract sockets are cleaner; with filesystem sockets, the file remains after the socket is closed (or the daemon crashes) and has to be removed separately (opening a possible race condition), and a directory with proper permissions is required.

The ifdefs handle the portability issue.

jnpkrn:
Closing in favor of PR #192.
See also the confirmation: [ClusterLabs](http://oss.clusterlabs.org/pipermail/developers/2016-March/000166.html)/[gmane](http://article.gmane.org/gmane.comp.clustering.clusterlabs.devel/127).

PR #188, opened 2016-02-29
Manual and static analysis -> fixes

jnpkrn:


jnpkrn:
Hm,

```
lt-ipc.test: ipcs.c:354: resend_event_notifications: Assertion `c->outstanding_notifiers <= sizeof(bytes)' failed.
```

which likely means also the original code has been slightly broken, but
it was never intentionally escalated as is now :-/

jnpkrn:
Note that I am able to reproduce the issue with

```
#define MAX_RECV_MSGS 395
```

(not 100% reliably, though).

Trying even more, although there's a suspicion that the sending side should
have the buffer `MAX_RECV_MSGS + X` where `X` is to be investigated.

jnpkrn:
Reproduced also with

```
#define MAX_RECV_MSGS 445
```

and with

```
#define MAX_RECV_MSGS 50
[...]
resend_event_notifications(struct qb_ipcs_connection *c)
                                                       
       ssize_t res = 0;                                
       static char bytes[MAX_RECV_MSGS * 2];           
```

Any idea what is going on and what's the reliable upper bound?

chrissie-c:
Closing this unmerged. While some of it is good, plenty is unfinished or dodgy and the decent bits are not with the effort of teasing them apart.

PR #191, opened 2016-03-01
tests: refactor test case defs using versatile add_tcase macro

jnpkrn:
This reduces repeated code significantly, and allows for easier
supervision of what's being grouped to the suites + possibly what
timeouts apply.


jnpkrn:
Forgot to add a note to the commit message that some artificial test case
identifiers got changed so they now follow 1:1 the function that is being
run for the case at hand without "test_" prefix.

PR #192, opened 2016-03-02
Add Hurd support

jnpkrn:
Again (see PR #187, which is superseded with this PR, if I am not mistaken), I am just a messenger.
Original available at [clusterlabs](http://oss.clusterlabs.org/pipermail/developers/2016-March/000164.html) (or [gmane](http://article.gmane.org/gmane.comp.clustering.clusterlabs.devel/125)).

```
  * configure.ac: Define QB_GNU

  * lib/log_thread.c: Replace second argument of
  qb_log_thread_priority_set(): logt_sched_param.sched_priority by 0
  when not supported by the OS.

  * lib/util.c: Fall back to CLOCK_REALTIME in clock_getres() if
    CLOCK_MONOTONIC fails.
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -173,6 +173,18 @@ AC_COMPILE_IFELSE([AC_LANG_PROGRAM(
                  )
 
 
+AC_MSG_CHECKING(for a working clock_getres(CLOCK_MONOTONIC, &ts))
+AC_RUN_IFELSE([AC_LANG_PROGRAM(
+[[#include <time.h>]],
+[[struct timespec ts; if(clock_getres(CLOCK_MONOTONIC, &ts)) return -1;]])],
+                    [
+                      AC_MSG_RESULT([yes])
+                      AC_DEFINE_UNQUOTED([HAVE_CLOCK_GETRES_MONOTONIC], 1, [Define to 1 if clock_getres(CLOCK_MONOTONIC, &ts) works])
+                    ],
+                    [
+                      AC_MSG_RESULT([no])
+                    ]
+                 )
 AC_MSG_CHECKING(for MSG_NOSIGNAL)
 AC_TRY_COMPILE([#include <sys/socket.h>],
 	       [ int f = MSG_NOSIGNAL; ],
@@ -334,6 +346,11 @@ case "$host_os" in
 				   [Compiling for Solaris platform])
 		CP=rsync
 		AC_MSG_RESULT([Solaris])
+	;;
+	*gnu*)
+		AC_DEFINE_UNQUOTED([QB_GNU], [1],
+				   [Compiling for GNU/Hurd platform])
+		AC_MSG_RESULT([GNU])
 	;;
 	*)
 		AC_MSG_ERROR([Unsupported OS? hmmmm])
--- a/lib/log_thread.c
+++ b/lib/log_thread.c
@@ -164,7 +164,11 @@ qb_log_thread_start(void)
 
 	if (logt_sched_param_queued) {
 		res = qb_log_thread_priority_set(logt_sched_policy,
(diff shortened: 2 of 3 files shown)


jnpkrn:
Some more comments:
- at least `lib/util.c` change doesn't conform to the surrounding coding style
- one would guess that failing `clock_getres(CLOCK_MONOTONIC, &ts)` is a property of the system, why not to figure this out at build time in `configure.ac`?

gnu-srs:
See the ML.  An updated patch is available.

gnu-srs:
ping, 1.0 is now released.

jnpkrn:
On 08/04/16 00:12 -0700, gnu-srs wrote:
> ping, 1.0 is now released.

It would be helpful if you shared a little bit of HOWTO with us
about GNU/Hurd system tryout.  Do you run it virtualized, under
KVM perhaps?  What's the easiest way?  Shall we start at
https://www.debian.org/ports/hurd/hurd-cd or is there
something more convenient for such purposes?

-- 
Jan (Poki)

chrissie-c:
To be honest, I've no real intention of testing Hurd myself - I'm mainly concerned that the patches don't break any other platforms. Testing on Hurd can be done by the people that want it

Not that I'm stopping you, of course ;-)

ChristophBerg:
Fwiw, the libqb Debian package contains a patch that looks similar to what is proposed for inclusion here: http://anonscm.debian.org/cgit/debian-ha/libqb.git/tree/debian/patches/hurd-support.patch

The buildd status of the just-uploaded 1.0 package looks fine as well: https://buildd.debian.org/status/package.php?p=libqb
(The x32 failure there is new, though, but likely unrelated as the patch was there before.)

chrissie-c:
So do you need the clock parts of the patch for hurd? I can't see those in the Debian patch you referenced? I'm quite amenable to merging that one if it does what you need.

gnu-srs:
On Wed, 2016-06-22 at 04:13 -0700, Christoph Berg wrote:
> Fwiw, the libqb Debian package contains a patch that looks similar to what is
> proposed for inclusion here: http://anonscm.debian.org/cgit/debian-
> ha/libqb.git/tree/debian/patches/hurd-support.patch

No they are not similar. In fact I'me reopening the Debian bug with the patch in
upstream #192.

> The buildd status of the just-uploaded 1.0 package looks fine as well:
> https://buildd.debian.org/status/package.php?p=libqb

With the old patch only 4 of 8 tests pass, and with the new patch 7 of 8 tests
pass. Only ipc.test remains. As written in #192 I'm working on a SCM_CREDS
solution for GNU/Hurd and GNU/kFreeBSD for that test, currently a WIP.
FYI: A test suite failure is made non-critical for the Debiaqn package to build.

gnu-srs:
On Tue, 2016-06-28 at 05:53 -0700, Chrissie Caulfield wrote:
> So do you need the clock parts of the patch for hurd? I can't see those in the
> Debian patch you referenced? I'm quite amenable to merging that one if it does
> what you need.

Yes, the clock test in configure.ac is needed. GNU/Hurd does not support the
call clock_getres(CLOCK_MONOTONIC, &ts) yet, therefore the test. The fallback
solution is to call clock_getres(CLOCK_REALTIME, &ts) instead.

chrissie-c:
I committed this prematurely but it wasn't causing any trouble in testing so I'm happy to leave it in.

gnu-srs:
On Tue, 2016-10-18 at 07:04 -0700, Chrissie Caulfield wrote:
> I committed this prematurely but it wasn't causing any trouble in testing so
> I'm happy to leave it in.
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub, or mute the thread.

Thanks!

PR #194, opened 2016-03-03
Memleak fixes

jnpkrn:
WIP, do not pull yet.

### DIFF
--- a/lib/ipc_socket.c
+++ b/lib/ipc_socket.c
@@ -329,6 +329,9 @@ _finish_connecting(struct qb_ipc_one_way *one_way)
 		return error;
 	}
 
+	/* Beside disposing no longer needed value, this also signals that
+	   we are done with connect-on-send arrangement at the server side
+	   (i.e. for response and event channels). */
 	free(one_way->u.us.sock_name);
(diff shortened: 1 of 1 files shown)


jnpkrn:
Actually I found another instance of missing `free(base_name);` and also
moved `sock_name` free'ing rather to `ESTABLISHED || ACTIVE` that feels
a right place for that.  Once being head down to that part, I've also
added some clarification to something that IMHO should have been clarified
for ages (magic constants are bad, alright?).

jnpkrn:
Also, the reason for CI failure is clear: their 64 MB sized /dev/shm gets full easily.

chrissie-c:
We need to look into the CI and test failures. If they fail randomly like this then they are worse than useless.

chrissie-c:
But the patch is good, I'll include it in the upcoming release. Thanks!

jnpkrn:
Yes, I've been meaning to add a parallel run to the test suite with
abstract sockets kill-switch enabled.

Regarding the intermittent failures when /dev/shm is full, we should
at least lean towards better reporting of that condition in libqb
itself -- there's no guarantees this won't happen in production use.

PR #214, opened 2016-06-16
CI: make Travis invoke also rpmlint for good measure

jnpkrn:
See also container-based infrastructure vs. rpmlint's setuid/setgid:

```
https://github.com/travis-ci/apt-package-whitelist/pull/2109
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,12 @@ compiler:
         - clang
 addons:
         apt:
+                sources:
+                        - debian-sid
                 packages:
                         - check
                         - splint
+                        - rpmlint
                         # for "make rpm"
                         - doxygen
                         - rpm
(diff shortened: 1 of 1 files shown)


jnpkrn:
For why this is failing, see:
https://github.com/travis-ci/travis-ci/issues/4838#issuecomment-226753908

Briefly: `rpmlint` is only available in `debian-sid` whitelisted source
(in Ubuntu proper, it's available since `wily`, 15.10:
http://packages.ubuntu.com/wily/rpmlint), but unfortunately enabling it,
conflict on `locales_2.22-11_all.deb` package will arise.
It would be nice to have the possibility to use that source just for
explicitly enumerated packages, which is currently not the case.

Alternatively, we could move from container-based infrastructure,
but as minor check as rpmlint likely does not justify slower build
run spin-ups.

Issue #215, opened 2016-06-17
[Problem] libqb does not seem to support a format of "%z".


HideoYamauchi:
Hi All,

The output of the log of Pacemaker was changed by the next correction.
 * https://github.com/ClusterLabs/pacemaker/commit/fc87717a91fa8c5011c184c1ea3ce551b650037d#diff-985358aeb562c6ddf99399a97c1424ec

When I appoint PCMK_blackbox=yes, Pacemaker fails in start.
It seems to have a problem that libqb does not support %z format.

Best Regards,
Hideo Yamauchi.

chrissie-c:
Proposed patch in PR#216

HideoYamauchi:
Hi Christine,

Okay!
I test it on Monday.

Best Regards,
Hideo Yamaucuhi.

sammcj:
Until resolved this likely blocks Redhat bug [1299968 - Rebase libqb to 1.0 for RHEL7.3](https://bugzilla.redhat.com/show_bug.cgi?id=1299968#c11)

HideoYamauchi:
Hi Christine,

I confirmed that your correction solved a problem.
 * https://github.com/ClusterLabs/libqb/pull/216

Best Regards,
Hideo Yamauchi.

chrissie-c:
Thanks for testing :)

chrissie-c:
Added as commit 22ac41108ec3ad273905c67ac5ef5be01bb0f268

HideoYamauchi:
Hi Christine,

Thanks!

kgaillot:
Pacemaker has been updated to avoid using %z as of commit 380c1bb, which will be part of the 1.1.15 release. So, no released version of Pacemaker will have a problem with older versions of corosync without this patch.

chrissie-c:
On 21/06/16 01:17, Ken Gaillot wrote:
> Pacemaker has been updated to avoid using %z as of commit 380c1bb, which
> will be part of the 1.1.15 release. So, no released version of Pacemaker
> will have a problem with older versions of corosync without this patch.

Thanks Ken

Chrissie

HideoYamauchi:
Hi Ken,

All right!

Mant thanks!
Hideo Yamauchi.

sammcj:
Just taking a moment here to say thank you to the ClusterLabs and RHEL communities, both have been so fantastic to work with and so many people really care about the quality and progression of even little issues like this. Thanks everyone 👍

PR #216, opened 2016-06-17
log: Add missing z,j, & t types to the logger

chrissie-c:
For problem #215

### DIFF
--- a/lib/log_format.c
+++ b/lib/log_format.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2011 Red Hat, Inc.
+ * Copyright (C) 2011,2016 Red Hat, Inc.
  *
  * All rights reserved.
  *
@@ -536,6 +536,33 @@ reprocess:
 				format++;
 			}
 			goto reprocess;
+		case 'z':
+			format++;
+			if (sizeof(size_t) == sizeof(long long)) {
+				type_longlong = QB_TRUE;
+			}
+			else {
+				type_long = QB_TRUE;
+			}
+			goto reprocess;
+		case 't':
+ 			format++;
+			if (sizeof(ptrdiff_t) == sizeof(long long)) {
+				type_longlong = QB_TRUE;
+			}
+			else {
+				type_long = QB_TRUE;
+			}
+			goto reprocess;
+		case 'j':
+ 			format++;
+			if (sizeof(intmax_t) == sizeof(long long)) {
+				type_longlong = QB_TRUE;
+			}
+			else {
+				type_long = QB_TRUE;
+			}
+			goto reprocess;
 		case 'd': /* int argument */
 		case 'i': /* int argument */
 		case 'o': /* unsigned int argument */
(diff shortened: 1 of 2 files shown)


lib/log_format.c
```diff
@@ -536,6 +536,33 @@ qb_vsnprintf_serialize(char *serialize, size_t max_len,
...
+			format++;
+			if (sizeof(size_t) == sizeof(long long)) {
+				type_longlong = QB_TRUE;
+			}
+			else {
+				type_long = QB_TRUE;
+			}
+			goto reprocess;
+		case 't':
+ 			format++;
```
jnpkrn: Please, do not mix spaces and tabs.
`git diff` (or perhaps `git diff --color`) should expose that visually
(using `color.diff.whitespace`) by default.

lib/log_format.c
```diff
@@ -536,6 +536,33 @@ qb_vsnprintf_serialize(char *serialize, size_t max_len,
 				format++;
 			}
 			goto reprocess;
+		case 'z':
+			format++;
+			if (sizeof(size_t) == sizeof(long long)) {
+				type_longlong = QB_TRUE;
+			}
+			else {
```
jnpkrn: This is not common style/indenting for follow-up `else` statements in libqb.

chrissie-c:
TBH the space/tab thing is a bit random all over even that file, but I've tidied it up and fixed the else format.

lib/log_format.c
```diff
@@ -737,6 +761,35 @@ qb_vsnprintf_deserialize(char *string, size_t str_len, const char *buf)
 				type_longlong = QB_TRUE;
 			}
 			goto reprocess;
+		case 'z':
+			fmt[fmt_pos++] = *format;
+			format++;
+			if (sizeof(size_t) == sizeof(long long)) {
+				type_long = QB_FALSE;
```
jnpkrn: `QB_FALSE` assignments for `'z'` are superfluous (unlike in `'l'` case).
At least if we don't want to care about behavior that is undefined anyway :-)

jnpkrn:
There's one more cosmetic discrepancy.
Isn't it better to `--amend`/merge and `git push --force` to keep the history compact?

chrissie-c:
git merging makes me break out in a cold sweat.

kgaillot:
This looks good to me, and is a good idea.

Unfortunately, it doesn't really help us in pacemaker, since we have to stay compatible with older libqb versions. We'll have to take all these formatters out and cast to long long instead. (That seems better than defining some ugly constants for format and cast depending on libqb support.)

chrissie-c:
Added as commit 22ac41108ec3ad273905c67ac5ef5be01bb0f268

chrissie-c:
On 17/06/16 22:22, Ken Gaillot wrote:
> This looks good to me, and is a good idea.
> 
> Unfortunately, it doesn't really help us in pacemaker, since we have to
> stay compatible with older libqb versions. We'll have to take all these
> formatters out and cast to long long instead. (That seems better than
> defining some ugly constants for format and cast depending on libqb
> support.)
>

Thanks. In that case I'll leave this in git until the next minor release
of libqb.

Chrissie

PR #217, opened 2016-06-17
Low: log: check for appropriate space when serializing a char

jnpkrn:
... where appropriate space is measured for, surprisingly, a char,
not for an int.  Note that's also the actual type used for both
de-/serializing, so there's no conflict.

Also bother to explain why, now surprisingly for real, an unsigned int
is scraped out from va_list (akin to to STDARG(3)).


chrissie-c:
Merged, thanks.

PR #218, opened 2016-06-21
docs: qbdefs.h: description must directly follow @file

wferi:
If we want to see it again in the man page NAME section, where it can
be indexed by apropos or whatis.


wferi:
Branch rebased onto current master.

Issue #229, opened 2016-10-19
regression: [6bd3f0865a713e641e37c98a391696a5383abd4e] Add Hurd support 


fabbione:
above commit introduces a regression and corosync master is unable to start.

CPU is spinning 100% in:

18315 sched_setscheduler(18316, SCHED_RR, { 0 }) = -1 EINVAL (Invalid argument)
18315 futex(0x7fb56aebdca0, FUTEX_WAKE_PRIVATE, 1) = 1
18315 futex(0x7fb566ccb9d0, FUTEX_WAIT, 18316, NULL <unfinished ...>
18316 <... futex resumed> )             = 0
18316 --- SIGSEGV {si_signo=SIGSEGV, si_code=SEGV_MAPERR, si_addr=0} ---
18316 write(7, "\v\0\0\0", 4)           = 4
18316 rt_sigreturn()                    = 0
18316 --- SIGSEGV {si_signo=SIGSEGV, si_code=SEGV_MAPERR, si_addr=0} ---
18316 write(7, "\v\0\0\0", 4)           = 4
18316 rt_sigreturn()                    = 0
18316 --- SIGSEGV {si_signo=SIGSEGV, si_code=SEGV_MAPERR, si_addr=0} ---
18316 write(7, "\v\0\0\0", 4)           = 4
18316 rt_sigreturn()                    = 0

fabbione:
diff --git a/lib/log_thread.c b/lib/log_thread.c
index a0cd25c..c67c294 100644
--- a/lib/log_thread.c
+++ b/lib/log_thread.c
@@ -164,11 +164,7 @@ qb_log_thread_start(void)
 
        if (logt_sched_param_queued) {
                res = qb_log_thread_priority_set(logt_sched_policy,
-#if defined(HAVE_PTHREAD_SETSCHEDPARAM) && defined(HAVE_SCHED_GET_PRIORITY_MAX)
                                                 logt_sched_param.sched_priority);
-#else
-                                                0);
-#endif
                if (res != 0) {
                        goto cleanup_pthread;
                }

this is the problematic chunk of code. HAVE_SCHED_GET_PRIORITY_MAX is never set. configure.ac doesn´t even check for it and the code does not use sched_get_priority_max anywhere anyway.

the check for HAVE_PTHREAD_SETSCHEDPARAM is redundant because qb_log_thread_priority_set is already wrapped in #if defined HAVE_......

jnpkrn:
[...]

PR #230, opened 2016-10-20
Med: log_thread: logt_wthread_lock is vital for logging thread

jnpkrn:
This fixes issue with would-fail-if-applied-to-thread-right-away
qb_log_thread_priority_set invocation when logging thread doesn't
exist yet, which will arrange for calling itself at the time of
thread's birth that is the moment it will actually fail.
In this + lock-could-not-have-been-initialized corner cases, the
already running thread would proceed as allowed by error condition
handling in the main thread, trying to dereference uninitialized
(or outdated) pointer to the lock at hand, resulting in segfault.

Also include the test that would have been caught that (we use the
fact that it doesn't matter whether setting of the scheduler parameters
fails due to bad input or just because of lack of privileges as it's
the failure at the right moment that is of our interest).

See also:
https://github.com/ClusterLabs/libqb/issues/229

### DIFF
--- a/lib/log_thread.c
+++ b/lib/log_thread.c
@@ -146,6 +146,7 @@ int32_t
 qb_log_thread_start(void)
 {
 	int res;
+	qb_thread_lock_t *wthread_lock;
 
 	if (wthread_active) {
 		return 0;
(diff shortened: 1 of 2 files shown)


chrissie-c:
Good catch. thanks.

PR #231, opened 2016-11-03
Unlink or truncate (as a fallback) files when shm IPC client terminates connection forcibly

jnpkrn:


lib/ringbuffer_helper.c
```diff
@@ -306,3 +306,86 @@ qb_rb_sem_create(struct qb_ringbuffer_s * rb, uint32_t flags)
...
+#  else
+#    define RB_DIR_RO_FLAGS1 O_RDONLY|O_DIRECTORY
+#  endif
+#  ifndef O_PATH
+#    define RB_DIR_RO_FLAGS RB_DIR_RO_FLAGS1
+#  else
+#    define RB_DIR_RO_FLAGS RB_DIR_RO_FLAGS1|O_PATH
+#  endif
+		char *data_path = rb->shared_hdr->data_path;
+		char *sep = strrchr(data_path, '/');
```
chrissie-c: I'm not keen on all those #defines in the middle of a function. Can you move them nearer the top of the file with an appropriate comment please?

jnpkrn: See https://github.com/ClusterLabs/libqb/pull/231/commits/15591922346ad485eb054f4c611b226dc6315393#diff-55541c07268e810475b8e7d661314605R311
Thanks for making me review the actual commits -- I've discovered that I had accidentally squashed two commits into what was the last one.  Now they are split again.

Issue #220, opened 2016-06-23
Please sign the uncompressed release tarball or each compressed one


wferi:
It's great that you now distribute proper signed releases, it really makes packaging your software much easier. However, the Debian helper scripts aren't prepared to validate your two-step scheme (checking the detached signature of the sha256sum file, then checking the hash of the compressed tarball against that). While we could implement this on our part, maybe you'd agree to switch to a more usual scheme, that is, to signing the uncompressed tarball or signing each compressed tarball separately. Both avoid (but does not forbid) the sha256sum files. The latter method requires multiple signatures, but is the simplest to handle on the receiver side.
Thanks for your consideration.

jnpkrn:
Perhaps the most reasonable way out is to say bye to `tar.gz` and distribute solely `tar.xz` + it's signature. At worst, downstream would be forced to recompress (still verifiable) tarball, but note that `tar` supports `xz` pass-through as of 7+ years ago.

chrissie-c:
I don't see any real problem with signing both compressed tarballs if that will help packagers.

wferi:
Yes please, that would be great! And then we wouldn't need the SHA256 files anymore.

wferi:
Hi, did you intend to sign the compressed tarballs of 1.0.1? :)

jnpkrn:
@wferi, regarding #236, are `.asc` signatures OK?

wferi:
Yes, @jnpkrn, ASCII armored signatures would be great. Just like raw ones. It does not matter, really.

wferi:
Hi, why not check the machinery on the fresh 1.0.2 release? :)

chrissie-c:
Should have closed this with 1.0.2!

PR #236, opened 2016-11-25
Overhaul build-aux/release.mk

jnpkrn:
Ultimate goal is to tackle #220.


jnpkrn:
Note to self: if this gets pulled in, update
http://clusterlabs.org/wiki/Point_Release_Checklist#Release_process
to make `gpgsignkey` mandatory.

jnpkrn:
Closing this PR as this was (in fact unknowingly, as I've forgotten about
this one) with #260 and later post-adjustments (e.g. #262).
There's at least one more thing that needs to be fixed in release.mk,
so I'll add some ideas from here, if any.

Issue #234, opened 2016-11-23
broken `test_ipc_max_dgram_size` test needs to be reviewed


jnpkrn:
https://travis-ci.org/ClusterLabs/libqb/jobs/178242766#L2722

```
../../tests/check_ipc.c:1498:F:ipc_max_dgram_size:test_ipc_max_dgram_size:0: Assertion 'init==try' failed: init==331264, try==425728
```

...triggered intermittently only with clang (3.4), upon unrelated change.

jnpkrn:
```
Build image provisioning date and time
Thu Feb  5 15:09:33 UTC 2015

Operating System Details

Distributor ID:	Ubuntu
Description:	Ubuntu 12.04.5 LTS
Release:	12.04
Codename:	precise

Linux Version
3.13.0-29-generic

Cookbooks Version
a68419e https://github.com/travis-ci/travis-cookbooks/tree/a68419e
```

jnpkrn:
Verified this is indeed intermittent, the above link now points to a restarted run, which passed (note that the offset of the respective line is `+3`, unfortunately I haven't grabbed these 3 extraneous lines when it was possible, which might have shed more light into this, supposing they were related error messages).

jnpkrn:
One of the possibilities that are hard to rule out is that parallel
matrix builds (e.g., multiple compilers) share the same /dev/shm
path (containers set up like that?) and it doesn't play very well
in some rare circumstances as similar pseudorandom paths are being
accessed...

chrissie-c:
Very odd. I'm not going to worry about it short-term, though it would be useful to know how the test systems are set up. Can we reproduce it with clang ourselves?

jnpkrn:
No cycles to spend on trying to reproduce that though we are now aware about this inclination in Travis CI so we'll have at least some clues when/if this recidivate.

jnpkrn:
Some archeology:
- there was once a magic issue 155 of travis-ci/travis-cookbooks,
  not available anymore (Issues got disabled for that project)
  . Angus once added this as a workaround:
    https://github.com/ClusterLabs/libqb/commit/ffdc2d519ab7f30e7d16490b37fe2fa08be00c3d#diff-354f30a63fb0907d4ad57269548329e3R6
  . which we dropped once we migrated from "legacy infrastructure"
    of Travis CI:
[...]

PR #240, opened 2016-12-07
Mostly Autoconf cleanups

wferi:
A couple of small things I noticed during packaging the 1.0.1 release. They don't make a big difference for a small project like libqb, but I decided to offer them for the sake of consistency and future-proofing. 
Especially for pthreads, where I switched to the standard macro from the Autoconf Archive.
I couldn't fully test this on systems where `gethostbyname` or the socket functions aren't in libc, so some corrections might be needed if you support them for real.
Please consider incorporating these patches.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -102,12 +102,7 @@ case $exec_prefix in
 esac
 
 # Checks for libraries.
-dnl librt from glibc NEEDs libpthread
-dnl so. if test for libpthread after librt
-dnl it will always be "none needed", but it is not true
-dnl when linking libraries. Looks like a bug.
-AC_SEARCH_LIBS([pthread_create], [pthread])
-AC_SEARCH_LIBS([mq_open], [rt])
+AX_PTHREAD
 AX_SAVE_FLAGS
 AC_SEARCH_LIBS([dlopen],[dl],,[AC_MSG_ERROR([cannot find dlopen() function])])
 AC_SUBST([dlopen_LIBS],[$LIBS])
@@ -201,13 +196,17 @@ AC_CHECK_FUNCS([alarm clock_gettime ftruncate gettimeofday \
 		strchr strrchr strdup strstr strcasecmp \
 		poll epoll_create epoll_create1 kqueue \
 		random rand getrlimit sysconf \
-		pthread_spin_lock pthread_setschedparam \
-                pthread_mutexattr_setpshared \
-                pthread_condattr_setpshared \
-		sem_timedwait semtimedop \
 		getpeerucred getpeereid \
 		openat unlinkat])
 
+AX_SAVE_FLAGS
+CFLAGS="$CFLAGS $PTHREAD_CFLAGS"
+LIBS="$LIBS $PTHREAD_LIBS"
+AC_CHECK_FUNCS([pthread_spin_lock pthread_setschedparam \
+		pthread_mutexattr_setpshared \
+		pthread_condattr_setpshared \
+		sem_timedwait semtimedop])
+AX_RESTORE_FLAGS
 
 # Checks for defined macros
 AC_MSG_CHECKING(for MSG_NOSIGNAL)
--- a/configure.ac
+++ b/configure.ac
@@ -111,7 +111,10 @@ AX_SAVE_FLAGS
 AC_SEARCH_LIBS([socket], [socket],,[AC_MSG_ERROR([cannot find socket() function])])
 AC_SUBST([socket_LIBS],[$LIBS])
 AX_RESTORE_FLAGS
-AC_SEARCH_LIBS([gethostbyname], [nsl])
+AX_SAVE_FLAGS
+AC_SEARCH_LIBS([gethostbyname], [nsl],,[AC_MSG_ERROR([cannot find gethostbyname() function])])
+AC_SUBST([nsl_LIBS],[$LIBS])
+AX_RESTORE_FLAGS
 
 # look for testing harness "check"
 PKG_CHECK_MODULES([CHECK], [check >= 0.9.4],[with_check=yes],[with_check=no])
--- a/configure.ac
+++ b/configure.ac
@@ -107,7 +107,10 @@ AX_SAVE_FLAGS
 AC_SEARCH_LIBS([dlopen],[dl],,[AC_MSG_ERROR([cannot find dlopen() function])])
 AC_SUBST([dlopen_LIBS],[$LIBS])
 AX_RESTORE_FLAGS
-AC_SEARCH_LIBS([socket], [socket])
+AX_SAVE_FLAGS
+AC_SEARCH_LIBS([socket], [socket],,[AC_MSG_ERROR([cannot find socket() function])])
+AC_SUBST([socket_LIBS],[$LIBS])
+AX_RESTORE_FLAGS
 AC_SEARCH_LIBS([gethostbyname], [nsl])
 
 # look for testing harness "check"
--- a/configure.ac
+++ b/configure.ac
@@ -452,33 +452,27 @@ LINT_FLAGS="-weak -D__gnuc_va_list=va_list -D__attribute\(x\)= \
 
 # local options
 AC_ARG_ENABLE([ansi],
-	[  --enable-ansi           : force to build with ANSI standards. ],
-	[ default="no" ])
+  [AS_HELP_STRING([--enable-ansi],[force to build with ANSI standards])])
 
 AC_ARG_ENABLE([fatal-warnings],
-	[  --enable-fatal-warnings : enable fatal warnings. ],
-	[ default="no" ])
+  [AS_HELP_STRING([--enable-fatal-warnings],[enable fatal warnings])])
 
 AC_ARG_ENABLE([debug],
-	[  --enable-debug          : enable debug build. ],
-	[ default="no" ])
+  [AS_HELP_STRING([--enable-debug],[enable debug build])])
 
 AC_ARG_ENABLE([coverage],
-	[  --enable-coverage       : coverage analysis of the codebase. ],
-	[ default="no" ])
+  [AS_HELP_STRING([--enable-coverage],[coverage analysis of the codebase])])
 
 AC_ARG_ENABLE([slow-tests],
-	[  --enable-slow-tests     : build and run slow tests. ],
-	[ default="no" ])
+  [AS_HELP_STRING([--enable-slow-tests],[build and run slow tests])])
 
 if test x"$with_check" = xyes; then
 AC_ARG_ENABLE([syslog-tests],
-	[  --enable-syslog-tests   : build and run syslog tests. ],
-	[ default="no" ])
+  [AS_HELP_STRING([--enable-syslog-tests],[build and run syslog tests])])
 fi
 
 AC_ARG_WITH([socket-dir],
-	[  --with-socket-dir=DIR   : socket dir. ],
+  [AS_HELP_STRING([--with-socket-dir=DIR],[socket directory @<:@LOCALSTATEDIR/run@:>@])],
 	[ SOCKETDIR="$withval" ],
 	[ SOCKETDIR="$localstatedir/run" ])
 
--- a/configure.ac
+++ b/configure.ac
@@ -108,7 +108,10 @@ dnl it will always be "none needed", but it is not true
 dnl when linking libraries. Looks like a bug.
 AC_SEARCH_LIBS([pthread_create], [pthread])
 AC_SEARCH_LIBS([mq_open], [rt])
-AC_SEARCH_LIBS([dlopen], [dl])
+AX_SAVE_FLAGS
+AC_SEARCH_LIBS([dlopen],[dl],,[AC_MSG_ERROR([cannot find dlopen() function])])
+AC_SUBST([dlopen_LIBS],[$LIBS])
+AX_RESTORE_FLAGS
 AC_SEARCH_LIBS([socket], [socket])
 AC_SEARCH_LIBS([gethostbyname], [nsl])
 
--- a/examples/Makefile.am
+++ b/examples/Makefile.am
@@ -33,11 +33,11 @@ simplelog_LDADD = $(top_builddir)/lib/libqb.la
 
 tcpclient_SOURCES = tcpclient.c $(top_builddir)/include/qb/qbloop.h
 tcpclient_CPPFLAGS = -I$(top_builddir)/include -I$(top_srcdir)/include
-tcpclient_LDADD = $(top_builddir)/lib/libqb.la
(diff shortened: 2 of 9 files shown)


jnpkrn:
Minor nitpicks:
- could you please stick with common commit message prefixes
  and style that has been exercised lately
  (`build: [configure:]`, `Fix typos:`)?
- it's entirely fine to merge the single-change spelling fixes
- the first commit's message contains typo on its own (`defualt`)

configure.ac
```diff
@@ -466,10 +466,8 @@ AC_ARG_ENABLE([coverage],
 AC_ARG_ENABLE([slow-tests],
   [AS_HELP_STRING([--enable-slow-tests],[build and run slow tests])])
 
-if test x"$with_check" = xyes; then
 AC_ARG_ENABLE([syslog-tests],
   [AS_HELP_STRING([--enable-syslog-tests],[build and run syslog tests])])
```
jnpkrn: Can you advise as to whether this conditional is sensible, please?
My original intention was to lay an egg as carefully as possible because
libtool warns that `foo_LDFLAGS = -module` is not portable.

Any idea of having this portability tested in some minimalistic way
and turning "syslog-tests" on automatically if positive?

Or is it more a spurious warning of libtool?

jnpkrn: Hm, cannot reproduce that easily, it might have been on FreeBSD 10.2 VM.

jnpkrn: Reproduced on Linux box, but had to go back in history:

```
$ make distclean
$ git co -b syslog-test  642f74d
$ ./autogen.sh && ./configure --enable-syslog-tests && make check
...
*** Warning: Linking the executable log.test against the loadable module
*** _syslog_override.so is not portable!
...
```

Hard to say what resolved this.

wferi: ebcff547bc47a6a3c195f6eb9d839ffed60a7381 did. Looks like portability is a concern for installed artifacts only. According to [SO](http://stackoverflow.com/questions/7619607/how-portable-is-linking-executables-against-loadable-modules) and the included [link](http://stackoverflow.com/questions/2339679/what-are-the-differences-between-so-and-dylib-on-osx), only Mac OS X is affected.

jnpkrn: Yep, just have discovered that commit too; nice to have this positive
side effect :-)

In the new light, I would suggest dropping `--enable-syslog-tests`
unconditionally when the test suite is enabled as a whole.
I intend to do it subsequently on top of your PR, but feel free to drop
it right now without attempting to rectify anything related needlessly.

wferi:
I implemented your nits and reordered that patches. The last two (GitHub doesn't show them in ancestry order) may not be necessary if you intend to drop the `--enable-syslog-tests` option altogether. I can take them back if you prefer.

jnpkrn:
@wferi please have a look at #241 if it works for you

wferi:
#241 works for me perfectly, thanks.
Closing this.

jnpkrn:
On 08/12/16 13:30 -0800, wferi wrote:
> #241 works for me perfectly, thanks.
> Closing this.

I thank you for the exemplary cleanups, good work.

-- 
Jan (Poki)

configure.ac
```diff
@@ -108,7 +108,10 @@ dnl it will always be "none needed", but it is not true
 dnl when linking libraries. Looks like a bug.
 AC_SEARCH_LIBS([pthread_create], [pthread])
 AC_SEARCH_LIBS([mq_open], [rt])
-AC_SEARCH_LIBS([dlopen], [dl])
+AX_SAVE_FLAGS
+AC_SEARCH_LIBS([dlopen],[dl],,[AC_MSG_ERROR([cannot find dlopen() function])])
+AC_SUBST([dlopen_LIBS],[$LIBS])
+AX_RESTORE_FLAGS
```
jnpkrn: @wferi, discussed with @chrissie-c, she raised concerns about "keeping large(ish) chunks of other people's code in there just for simple things, it's just more stuff to keep up-to-date" and I must admit there's some grain of truth in it as all we really need here is just to back `LIBS`, nothing more, nothing less.  Would mind reopening this PR (if possible, I am not sure) and reflecting this comment, please?  Indeed, I can do it as a another follow up, but better to make history clean.

wferi: Right there you don't even need to save `LIBS`, because it's empty. It's a little more interesting around the pthread function tests, where we also manipulate `CFLAGS` (which might be set by the user). Sure, all uses of `AX_SAVE/RESTORE_FLAGS` could easily be replaced by custom code. I decided to use them in this pull request, though, mostly for demo.
The main point is that they save you from thinking globally when doing local changes: you gain modularity and composability. You are free to add more involved checks (possibly requiring other flag changes) inside the save/restore pair or reorder them at will. They also reduce the room for error (I might be biased, but a couple of weeks ago I found the insidious `CFLAGS="-I/usr/include/security $CLFAGS"` and `CPPFLAGS="$ac_save_CFLAGS"` in the same `configure.ac` — after far too much head scratching).
Concerning the other point, all three macro files are straight from the Autoconf Archive. They seldom need updates, and when they do, it's a simple replace. Another option is requiring the archive to be installed for building libqb from Git.

jnpkrn:
On 09/12/16 13:12 -0800, wferi wrote:
> wferi commented on this pull request.
> 
>> @@ -108,7 +108,10 @@ dnl it will always be "none needed", but it is not true
>  dnl when linking libraries. Looks like a bug.
>  AC_SEARCH_LIBS([pthread_create], [pthread])
>  AC_SEARCH_LIBS([mq_open], [rt])
> -AC_SEARCH_LIBS([dlopen], [dl])
> +AX_SAVE_FLAGS
> +AC_SEARCH_LIBS([dlopen],[dl],,[AC_MSG_ERROR([cannot find dlopen() function])])
> +AC_SUBST([dlopen_LIBS],[$LIBS])
> +AX_RESTORE_FLAGS
> 
> Right there you don't even need to save `LIBS`, because it's empty.
> It's a little more interesting around the pthread function tests,
> where we also manipulate `CFLAGS` (which might be set by the user).

Ah, next time, I will not forward anything semi-unconsciously (for
some reason, I went by just the easy cases).

> Sure, all uses of `AX_SAVE/RESTORE_FLAGS` could easily be replaced
> by custom code. I decided to use them in this pull request, though,
> mostly for demo.
> The main point is that they save you from thinking globally when
> doing local changes: you gain modularity and composability. You are
> free to add more involved checks (possibly requiring other flag
> changes) inside the save/restore pair or reorder them at will. They
> also reduce the room for error (I might be biased, but a couple of
> weeks ago I found the insidious `CFLAGS="-I/usr/include/security
> $CLFAGS"` and `CPPFLAGS="$ac_save_CFLAGS"` in the same
> `configure.ac` — after far too much head scratching).

Yes, the ability to offload some tasks to well-tested, self-contained
solution, is appealing.  No doubts about that.

> Concerning the other point, all three macro files are straight from
> the Autoconf Archive. They seldom need updates, and when they do,
> it's a simple replace. Another option is requiring the archive to be
> installed for building libqb from Git.

That would just complicate consumption, in-tree is OK I believe.

I guess I let Chrissie speak for herself.


Btw. if there's anything to be done about the commits, I would have
a tiny wish (feel free to ignore if the PR would eventually be
accepted as is):
"configure: restrict pthreads to where it's actually needed" commit
could mention that mq_open is no longer relevant beyond commit
70a9623a48a7cc30ec079aa2762a35a4b7fc7add.

Thanks for bearing with me (us).

-- 
Jan (Poki)

wferi:
Jan Pokorný <notifications@github.com> writes:

> "configure: restrict pthreads to where it's actually needed" commit
> could mention that mq_open is no longer relevant beyond commit
> 70a9623a48a7cc30ec079aa2762a35a4b7fc7add.

True; I added this note and dropped the last two commits reorganizing
--enable-syslog-tests, since you'll remove that anyway.

> Thanks for bearing with me (us).

Come on, I actually count on you being picky and careful.  I've only got
a very narrow view of the project and zero experience with maintaining it.
-- 
Regards,
Feri

wferi:
... and the Travis checks failed. On code which is exactly the same as 2908242 in #241. Is this failure genuine? Or maybe the `ipc.test` has some race? The HPPA buildd failed [differently](https://buildd.debian.org/status/fetch.php?pkg=libqb&arch=hppa&ver=1.0.1-1&stamp=1481130367). (Hurd and kFreeBSD also fail some tests.)

jnpkrn:
On 12/12/16 03:14 -0800, wferi wrote:
> ... and the Travis checks failed. On code which is exactly the same
> as 2908242 in #241. Is this failure genuine? Or maybe the `ipc.test`
> has some race?

Yes, I suspect something like this:
https://github.com/ClusterLabs/libqb/issues/234

> The HPPA buildd failed
> [differently](https://buildd.debian.org/status/fetch.php?pkg=libqb&arch=hppa&ver=1.0.1-1&stamp=1481130367).
> (Hurd and kFreeBSD also fail some tests.)

check_ipc.c:1007:E:ipc_stress_connections_shm:test_ipc_stress_connections_shm:0:
(after this point) Test timeout expired

^ just adjusting the timeout for that test might help here.

Anyway, updated #241 and it passed:
https://travis-ci.org/ClusterLabs/libqb/builds/183218749

-- 
Jan (Poki)

chrissie-c:
The timeout for the stress test has to be HUGE on some platforms. I think it's set to an hour or so even from *BSD

jnpkrn:
It now causes a following problem with EL6:
https://copr-be.cloud.fedoraproject.org/results/%40ClusterLabs/devel/epel-6-x86_64/00490031-libqb/build.log.gz

```
autoreconf: running: /usr/bin/autoconf
configure:19163: error: possibly undefined macro: AS_VAR_COPY
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
autoreconf: /usr/bin/autoconf failed with exit status: 1
```

Note that I already encountered something similar with pacemaker,
where solution was to use `AS_VAR_GET` instead:

https://github.com/ClusterLabs/pacemaker/commit/36f7196f0bf656f95d24ce826ca13b2d1c697dbe#diff-a5522b90382504dc530177f63ced5209R46

@wferi, if you are OK with such a change, I'll go ahead along with
resetting serial to minimal value (currently unsure if 0 or 1).

wferi:
@jnpkrn, what do you think about introducing a "legacy" M4 file like for example https://github.com/varnishcache/varnish-cache/blob/master/varnish-legacy.m4? Definitions of missing macros could be put there as needed in the future. We could start with something like
~~~
# For compatibility with autoconf < 2.63b
m4_ifndef([AS_VAR_COPY],
  [m4_define([AS_VAR_COPY],
     [AS_LITERAL_IF([$1[]$2], [$1=$$2], [eval $1=\$$2])])])
~~~
which is basically a conditionalized copy of the `AS_VAR_COPY` definition in `/usr/share/autoconf/m4sugar/m4sh.m4`.

jnpkrn:
On 18/12/16 02:01 -0800, wferi wrote:
> @jnpkrn, what do you think about introducing a "legacy" M4 file like

Sounds reasonable, though it needs to be tested if it would be indeed
the "fixed point" for ~el6 already or more and more macros would have
to be introduced like this, which is something we would rather avoid.

Will have a look early in 2017, holiday season for me till then
(in case you have it similar, enjoy).

-- 
Jan (Poki)

wferi:
Sure we can return to this next year.
I'm already on holiday.  Merry Christmas!
-- 
Feri

wferi:
Thanks for the merge, @jnpkrn.
How did you solve the problem of `AS_VAR_COPY` being undefined under EL6 in the end?

jnpkrn:
On 31/05/17 07:53 -0700, wferi wrote:
> How did you solve the problem of `AS_VAR_COPY` being undefined under
> EL6 in the end?

It rests on the back burner so far, will hopefully get to is soon.

-- 
Jan (Poki)

inouekazu:
> wferi commented on 31 May
> How did you solve the problem of AS_VAR_COPY being undefined under EL6 in the end?

I made a patch https://github.com/inouekazu/libqb/commit/4b3c4c7d447a5773d286c7890f3d4136fae33653. how is it?

wferi:
> I made a patch inouekazu/libqb@4b3c4c7. how is it?

Looks fine, but is it enough to fix the EL6 build, as @jnpkrn queried?

inouekazu:
> Looks fine, but is it enough to fix the EL6 build, as @jnpkrn queried?

Yes. I checked with RHEL 6.9.
```
[root@rhel69-1 ~]# cat /etc/redhat-release
Red Hat Enterprise Linux Server release 6.9 (Santiago)

[root@rhel69-1 ~]# git clone https://github.com/ClusterLabs/libqb
[root@rhel69-1 ~]# cd libqb

[root@rhel69-1 libqb]# vi m4/ax_compat.m4
# For compatibility with autoconf < 2.63b
m4_ifndef([AS_VAR_COPY],
  [AC_DEFUN([AS_VAR_COPY],
     [AS_LITERAL_IF([$1[]$2], [$1=$$2], [eval $1=\$$2])])])

[root@rhel69-1 libqb]# ./autogen.sh
(snip)
examples/Makefile.am: installing `build-aux/depcomp'
autoreconf: Leaving directory `.'
Now run ./configure and make

[root@rhel69-1 libqb]# ./configure
(snip)

libqb configuration:
  Version                  = 1.0.2.18-333f
  Prefix                   = /usr
  Executables              = ${exec_prefix}/sbin
  Man pages                = ${datarootdir}/man
  Doc dir                  = ${datarootdir}/doc/${PACKAGE_TARNAME}
  Libraries                = /usr/lib64
  Header files             = ${prefix}/include
  Arch-independent files   = ${datarootdir}
  State information        = /var
  System configuration     = /etc
  SOCKETDIR                = /var/run
  Features                 =  epoll gcc__sync attribute-section

libqb build info:
  Optimization             =
  Debug options            =
  Extra compiler warnings  =
  Env. defined CFLAG       = -g -O2
  Env. defined CPPFLAGS    =
  Env. defined LDFLAGS     =
  ANSI defined CPPFLAGS    =
  Coverage     CFLAGS      =
  Coverage     LDFLAGS     =
  Fatal War.   CFLAGS      =
  Final        CFLAGS      = -g -O2       -Wall -Wextra -Wunused -Wshadow -Wmissing-prototypes -Wmissing-declarations -Wstrict-prototypes -Wpointer-arith -Wwrite-strings -Wcast-align -Wbad-function-cast -Wmissing-format-attribute -Wfloat-equal -Wformat=2 -Woverlength-strings -Winit-self -Wuninitialized -Wunknown-pragmas -Wno-unused-parameter -Wno-format-nonliteral
  Final        CPPFLAGS    =
  Final        LDFLAGS     =
[root@rhel69-1 libqb]#
```

PR #242, opened 2016-12-12
CI: travis: fix du -> df and capture it also directly from test

jnpkrn:


chrissie-c:
What's the reasoning behind the "#if 1" in check_ipc.c?

jnpkrn:
It's conditionalizing part that perhaps someone will want to disable (e.g. downstream), for instance because the build/development environment also coincidentally exports `CI` environment variable and any byproduct is undesirable.  Once we isolate and fix the original #234 issue, the whole bigger block should be dropped, but this is an extra convenience toggle for others to use.

chrissie-c:
If you want to make it 'semi-conditional' then can you please make a #define for it and check for that. #if 1 is just unhelpful.

jnpkrn:
OK, I have a better solution, but that requires #241 already merged as it also touches `.travis.yml` and would generate a merge conflict needlessly for that.

jnpkrn:
I piggybacked also unification of custom.py tito extension with what's currently in pacemaker (for ease of maintenance).

PR #245, opened 2017-02-15
Med: loop: prevent direct recursion in handling particular signal

jnpkrn:
See http://lists.clusterlabs.org/pipermail/users/2017-February/005073.html

* * *

This is very PRELIMINARY solution (if it can be supposed to be one at all)
not intended to be merged but rather to open a discussion on how we can do
it better.


jnpkrn:
For that to actually work, also `signal_num` should likely be temporarily
unblocked, which is another a valid operation to do even from within
the signal handler itself per `signal(7)`.

jnpkrn:
See also #229.

PR #246, opened 2017-03-24
Allow compile-time configuration of filesystem sockets for Linux

chrissie-c:
Allow Linux users the option (at ./configure time) of abstract or filesystem sockets. While abstract sockets can be tidier (and get cleaned up automatically at process end), on some container systems the namespace is global, preventing multiple instances of the same application from running. This system also allows programs to communicate between containers (by sharing a filesystem) should that be needed.


chrissie-c:
Better patch coming.

PR #247, opened 2017-03-29
Remove runtime check for CLOCK_MONOTONIC

kraj:
Does not work during cross compile

use a compile-time check to detect whether the
monotonic clock is available. This check can run just fine when we are
cross-compiling.

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -179,9 +179,17 @@ AC_COMPILE_IFELSE([AC_LANG_PROGRAM(
                  )
 
 AC_MSG_CHECKING(for a working clock_getres(CLOCK_MONOTONIC, &ts))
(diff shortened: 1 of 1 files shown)


chrissie-c:
Thanks for the patch, but it doesn't work (on Linux or FreeBSD).

'int main() {' is unneeded as its added by autoconf, and the include of time.h should be in the same stanza as unistd.h. 

If you fix these things I'd be happy to pull it.

yann-morin-1998:
I've sent another, simpler patch to fix the same issue: #269.

chrissie-c:
#269 has been merged.

PR #248, opened 2017-04-27
Allow Linux to use filesystem sockets

chrissie-c:
This patch enables system-wide selection of filesystem or abstract sockets when running on Linux. The default is still to use abstract sockets.

Filesystem sockets are needed when running libqb servers inside some containers where the abstract socket name space is shared, so abstract socket names would clash.

PR #249, opened 2017-04-27
test: Fix random number generation in IPC tests

chrissie-c:
The sockets are named using a random() suffix in at attempt to isolate
concurrent test. However random() always returns the same random number
by design. So I've added srandom(getpid()) to the start of the
IPC tests to make the randomisation actually work.

### DIFF
--- a/tests/check_ipc.c
+++ b/tests/check_ipc.c
@@ -1583,6 +1583,8 @@ main(void)
 	Suite *s;
 	int32_t do_shm_tests = QB_TRUE;
 
+	srandom(getpid());
+
 	set_ipc_name("ipc_test");
 #ifdef DISABLE_IPC_SHM
 	do_shm_tests = QB_FALSE;


kgaillot:
FYI srandom()+random() isn't thread-safe (all threads will get the same numbers), but that's fine here as long as it's only used for IPC. You can use srandom_r() +random_r() for thread safety if you want it.

If an attacker would gain any advantage from knowing what will be chosen ahead of time (DoS by creating a file at that location?), or thread safety is desired, use something besides just the PID as the seed.

jnpkrn:
On 27/04/17 08:45 -0700, Chrissie Caulfield wrote:
> The sockets are named using a random() suffix in at attempt to isolate
> concurrent test. However random() always returns the same random number
> by design. So I've added srandom(getpid()) to the start of the
> IPC tests to make the randomisation actually work.

As for the purpose of the tests there's really no need to guard against
rogue collisions of any kind, just the accidental ones, we should be
fine with something like:

  snprintf(ipc_name, 256, "%s%lX%.4x", prefix, (long)getpid(),
           (int) ((long) time(NULL) % (0x10000)))

rather than playing a (previously un-)randomness theatre
(even this form is safer than previously generated deterministic
sequence).

For completeness, we are discussion just the control socket names, as
the SHM files are constructed by appending other variables (again pid,
for instance).

-- 
Jan (Poki)

jnpkrn:
BTW. this also coincides with a #248, which introduces abstract UNIX
sockets kill-switch because they are not properly namespaced in Linux
kernel (and hence in containers upon it).

And the Travis CI tests for libqb is run in such containers, so we'd
rather be careful :-)

-- 
Jan (Poki)

jnpkrn:
On 28/04/17 01:28 +0200, Jan Pokorný wrote:
> BTW. this also coincides with a #248, which introduces abstract UNIX
> sockets kill-switch because they are not properly namespaced in Linux
> kernel (and hence in containers upon it).
> 
> And the Travis CI tests for libqb is run in such containers, so we'd
> rather be careful :-)

Oh wait, the PIDs can also be too deterministic when comparing two
identical container runs.  Hmm, the next thing on my mind would be
then to sleep a second or two in case the socket to be created
already exists, and retry?

-- 
Jan (Poki)

chrissie-c:
I think we're getting a bit too involved it this. It's just the build-time test suite guarding against clashes that hardly ever happen! This is a bug that nobody has hit. How about srandom(time(NULL)) ?

jnpkrn:
On 28/04/17 00:05 -0700, Chrissie Caulfield wrote:
> I think we're getting a bit too involved it this. It's just the
> build-time test suite guarding against clashes that hardly ever
> happen! This is a bug that nobody has hit. How about
> srandom(time(NULL)) ?

From my perspective, the simplest solution is to drop playing with
pseudorandomness completely and just stick with varying environment
as suggested.  I can send a separate PR if desired.

-- 
Jan (Poki)

chrissie-c:
Please do :)

jnpkrn:
Done, #250

chrissie-c:
Closed in favour of #250

PR #252, opened 2017-05-18
ipc: fix compile warning on non-Linux platforms

chrissie-c:

### DIFF
--- a/lib/ipc_socket.c
+++ b/lib/ipc_socket.c
@@ -48,16 +48,16 @@ int use_filesystem_sockets(void)
 	static int filesystem_sockets = 0;
 
 	if (need_init) {
+#if defined(QB_LINUX) || defined(QB_CYGWIN)
 		struct stat buf;
 
-		need_init = 0;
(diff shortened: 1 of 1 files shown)


kgaillot:
looks good to me

jnpkrn:
I think that for platforms beside `QB_LINUX` and `QB_CYGWIN` it should
not even go through is-initiated part but return constant right away,
which the compiler can optimize out right away (perhaps with a hint
when the function is `inline`, which it is a good candidate for anyway).

Otherwise the fix is sound.

chrissie-c:
True, but it's not on the fast path.

jnpkrn:
I don't understand.

chrissie-c:
It's only called when the socket is created (at either end), so quite infrequently. Optimisation is not going to achieve much.

jnpkrn:
True.  Except for the test suite and similar "rapid" use cases.

chrissie-c:
The test suite has bigger problems ;-)

PR #253, opened 2017-05-19
build: follow-up on introducing custom m4 macros

jnpkrn:
Various "configure" commits by wferi recently introduced new
compat/custom m4 macro files in m4 directory, which itself was,
so far, assumed ephemeral (not strictly needed for reproducing
the build successfully, i.e., bits that can be completely purged
when cutting down the project files to the bone).  Apparently,
this assumption no longer holds so several places need to be
adapted.

Amonst others, m4 directory no longer needs to be reinsured in
autogen.sh, and special care must be taken with .gitignore
and maintainer-clean-local target of the main Makefile.

See also PR #240.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,11 @@ autom4te.cache/
 config.status
 configure
 .libs
-m4
+
+# ignore "libtoolized" m4 files, but keep our (custom-prefixed) ones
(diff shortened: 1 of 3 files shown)


chrissie-c:
Merged, thanks

PR #254, opened 2017-05-23
test: Make check_ipc tidy up after itself.

chrissie-c:

### DIFF
--- a/tests/check_ipc.c
+++ b/tests/check_ipc.c
@@ -265,10 +265,12 @@ s1_msg_process_fn(qb_ipcs_connection_t *c,
 		}
 
 	} else if (req_pt->id == IPC_MSG_REQ_SERVER_FAIL) {
+		qb_ipcs_destroy(s1);
 		exit(0);
 	} else if (req_pt->id == IPC_MSG_REQ_SERVER_DISCONNECT) {
 		multiple_connections = QB_FALSE;
 		qb_ipcs_disconnect(c);
+		qb_ipcs_destroy(s1);
 	}
 	return 0;
 }
@@ -417,9 +419,6 @@ run_ipc_server(void)
 	};
 	uint32_t max_size = MAX_MSG_SIZE;
 
-	qb_loop_signal_add(my_loop, QB_LOOP_HIGH, SIGTERM,
-			   NULL, exit_handler, &handle);
-
 	my_loop = qb_loop_create();
 
 	s1 = qb_ipcs_create(ipc_name, 4, ipc_type, &sh);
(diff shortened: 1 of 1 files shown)


tests/check_ipc.c
```diff
@@ -265,10 +265,12 @@ s1_msg_process_fn(qb_ipcs_connection_t *c,
 		}
 
 	} else if (req_pt->id == IPC_MSG_REQ_SERVER_FAIL) {
+		qb_ipcs_destroy(s1);
 		exit(0);
 	} else if (req_pt->id == IPC_MSG_REQ_SERVER_DISCONNECT) {
 		multiple_connections = QB_FALSE;
 		qb_ipcs_disconnect(c);
+		qb_ipcs_destroy(s1);
```
jnpkrn: Overall, the code logic is quite hard to follow, so I would encourage adding
comments along whenever there's something not entirely apparent.

In this very case, I have no clue why you wouldn't want to substitute
`disconnect` for `destroy`, but you put it in addition instead, why a superficial
look at the latter suggests the former is implied...

chrissie-c: Thanks. testing this on FreeBSD causes breakage so I think the qb_ipcs_destroy() calls cause the sockets to be closed when the client test is expecting them to be still there. I think I might have to revert to the manual unlink that I had to use into test_ipcc_truncate_when_unlink_fails_shm()

tests/check_ipc.c
```diff
@@ -429,15 +428,18 @@ run_ipc_server(void)
 		qb_ipcs_enforce_buffer_size(s1, max_size);
 	}
 	qb_ipcs_poll_handlers_set(s1, &ph);
+	qb_loop_signal_add(my_loop, QB_LOOP_HIGH, SIGTERM,
+			   NULL, exit_handler, &handle);
```
jnpkrn: Good catch that the former location made the very first invocation
of `qb_loop_signal_add` silently fail.

Can we have this fix as a separate commit, please?

tests/check_ipc.c
```diff
@@ -429,15 +428,18 @@ run_ipc_server(void)
...
 	res = qb_ipcs_run(s1);
 	ck_assert_int_eq(res, 0);
 
 	qb_loop_run(my_loop);
+	qb_ipcs_destroy(s1);
 	qb_log(LOG_DEBUG, "loop finished - done ...");
 }
 
-static int32_t
+static pid_t
```
jnpkrn: Again, good catch, again I'd vote for a separate commit :)

chrissie-c: Good idea. I'll put those two 'real' fixes in now. The cleanup ones rare being more elusive than I'd like

PR #257, opened 2017-05-26
Adds additional warnings

jonesmz:
This pull request reconfigures the warning settings in configure.ac.

The purpose is to leverage the compilers ability to analyze the code for potential pitfalls so that we can more easily address problems.

The specific changes will be documented in the diff.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -512,23 +512,29 @@ EXTRA_WARNINGS=""
 
 WARNLIST="
 	all
+	extra
+	unused
 	shadow
 	missing-prototypes
 	missing-declarations
+	suggest-attribute=noreturn
+	suggest-attribute=format
 	strict-prototypes
-	declaration-after-statement
 	pointer-arith
 	write-strings
 	cast-align
 	bad-function-cast
 	missing-format-attribute
+	float-equal
 	format=2
-	format-security
-	no-format-nonliteral
-	no-long-long
-	unsigned-char
-	gnu89-inline
-	no-strict-aliasing
+	format-signedness
+	shift-overflow=2
+	overlength-strings
+	redundent-decls
+	init-self
+	uninitialized
+	unknown-pragmas
+	no-unused-parameter
 	"
 
 for j in $WARNLIST; do
--- a/configure.ac
+++ b/configure.ac
@@ -535,6 +535,7 @@ WARNLIST="
 	uninitialized
 	unknown-pragmas
 	no-unused-parameter
+	no-format-nonliteral
 	"
 
 for j in $WARNLIST; do


configure.ac
```diff
@@ -512,23 +512,29 @@ EXTRA_WARNINGS=""
 
 WARNLIST="
 	all
+	extra
```
jonesmz: extra includes several warning types that significantly increase the ability for the compiler to warn about potential pitfalls.

configure.ac
```diff
@@ -512,23 +512,29 @@ EXTRA_WARNINGS=""
...
+	unused
```
jonesmz: -Wunused combined with -Wextra provides insight into several categories of code where a value is create but never used. This can encompass return values being ignored, or variables that are assigned to but never used. Generally situations like this imply that a function that can potentially return an error is having that possible error ignored.

configure.ac
```diff
@@ -512,23 +512,29 @@ EXTRA_WARNINGS=""
...
 	shadow
 	missing-prototypes
 	missing-declarations
+	suggest-attribute=noreturn
```
jonesmz: This flags functions that the compiler thinks never return as good candidates for the noreturn attribute. When the noreturn attribute is used in the headerfile / function declaration, it allows the compiler to optimize the call-site of the function in other translation units, such as not bothering to fill out the full framework for a function call, since the called function won't ever return back to it.

If this never triggers, that's great. Otherwise it'll tell of places where we can capture additional optimization.

configure.ac
```diff
@@ -512,23 +512,29 @@ EXTRA_WARNINGS=""
...
+	suggest-attribute=format
```
jonesmz: Like the above, this suggests call sites where the compiler thinks that a function is a good candidate for an attribute that will enable additional optimizations or features. In this case, it allows for a function accepting variable arguments, with the first argument being a string type, to be processed for security and/or compatibility concerns against a format string.

E.g.

foo(char *, ...)

if you call
foo("%s", (int)5);

gcc will emit a warning that the format doesn't match the parameter.

configure.ac
```diff
@@ -512,23 +512,29 @@ EXTRA_WARNINGS=""
...
 	strict-prototypes
-	declaration-after-statement
```
jonesmz: Obsolete with C99.

configure.ac
```diff
@@ -512,23 +512,29 @@ EXTRA_WARNINGS=""
...
 	pointer-arith
 	write-strings
 	cast-align
 	bad-function-cast
 	missing-format-attribute
+	float-equal
```
jonesmz: Directly comparing floating point values is very commonly a bug. Better is to compare +/- some epsilon.

configure.ac
```diff
@@ -512,23 +512,29 @@ EXTRA_WARNINGS=""
...
 	format=2
-	format-security
-	no-format-nonliteral
```
jonesmz: These two format types are subsumed by the -Wformat=2 and -Wextra options. No need to keep them.

configure.ac
```diff
@@ -512,23 +512,29 @@ EXTRA_WARNINGS=""
...
-	no-long-long
-	unsigned-char
```
jonesmz: This is not documented anywhere online that I can find. Gcc website, stackoverflow, google. If it's important, feel free to add it back in.

configure.ac
```diff
@@ -512,23 +512,29 @@ EXTRA_WARNINGS=""
...
-	gnu89-inline
```
jonesmz: Obsolete with C99

configure.ac
```diff
@@ -512,23 +512,29 @@ EXTRA_WARNINGS=""
...
-	no-strict-aliasing
+	format-signedness
```
jonesmz: Ensures that code isn't doing

print("%d", (unsigned int)5);
or
print("%u", (int)5);

configure.ac
```diff
@@ -512,23 +512,29 @@ EXTRA_WARNINGS=""
...
+	shift-overflow=2
```
jonesmz: Helps catch bit-shift operations invoking undefined behavior, or otherwise being very sneaky.

configure.ac
```diff
@@ -512,23 +512,29 @@ EXTRA_WARNINGS=""
...
+	overlength-strings
+	redundent-decls
+	init-self
+	uninitialized
```
jonesmz: These two catch bugs of the type

int foo()
{
    int i = i;
    return i;
}

configure.ac
```diff
@@ -512,23 +512,29 @@ EXTRA_WARNINGS=""
...
 	format=2
-	format-security
-	no-format-nonliteral
-	no-long-long
-	unsigned-char
-	gnu89-inline
-	no-strict-aliasing
+	format-signedness
+	shift-overflow=2
+	overlength-strings
```
jonesmz: Prevents defining raw string literals that exceed the length defined as the "minimum maximum" in the C standard. No strings in libqb exceed this, but to ensure no submissions that break future compatibility with platforms that have different support, this will catch that.

configure.ac
```diff
@@ -512,23 +512,29 @@ EXTRA_WARNINGS=""
...
+	redundent-decls
+	init-self
+	uninitialized
+	unknown-pragmas
+	no-unused-parameter
```
jonesmz: libqb has many hundreds of function parameters that are unused. 
This prevents them being warned about.

Perhaps in the future, when those parameters are used or suppressed, this can be turned back on.

jnpkrn:
Nice, and true, we gotta keep kicking in this direction unless we want
to witness a bit rot.  With quite a recent GCC, I've now received 266
warnings, that is, 253 warnings more than with current master branch
(from 13 originally only one is purposeful).

\o/ some maintenance work ahead :)

jonesmz:
Given my recent experience with contributing to the Corosync project (which has a similar list of contributors to libqb), I decided not to go out of my way to provide code fixes that were likely to be rejected.

If you can provide me with assurances that a patch to fix the newly introduced warnings will be reviewed and -- assuming no regressions -- merged within a reasonable time frame such as within 4 weeks, I'd be happy to contribute at least a large handful of fixes for the warnings that this pull request unhides, if not most / all of them depending on my availability.

However, I'm sadly not interested in contributing to projects that reject even minuscule and trivial to review patches. So please tell me now if I'll be wasting my time.

If you'd like to see fixes for these contributed by me, I'll very likely have many of them ready within the week for your review.

chrissie-c:
Thanks, they generate a good lot of warnings that we need to look closely at!

For some reason -Wformat-nonliteral seems to have got enabled (I'm not quite sure why it wasn't already activated by -Wformat=2 TBH) and it's generating warnings in log_format that we can't do anything about. With the addition of -Wno-format-nonliteral to disable those I'll be happy to pull this.

chrissie-c:
Merged, Thanks

PR #260, opened 2017-06-09
build: some cleanups to release.mk

jnpkrn:
In part a follow-up for PR #259.


jnpkrn:
Force-pushed a better version that also fixes the pre-existing race condition when fulfilling recursively the dependency chains (see 1st commit).

chrissie-c:
Merged, thanks!

PR #263, opened 2017-07-27
log: use fdatasync instead of fsync where possible

jnpkrn:
Using years-old benchmark attached to PostreSQL ML[1], I've observed
tiny bit more than double boost in speed when using fdatasync instead of
traditional fsync, on two Linux machines, each equipped with an SSD.
While the observation may be disputable (there are various
interpretations to what "synchronized I/O" actually means), by logical
extension of what the two are supposed to do, one can expect fdatasync
will perform no worse than fsync.  Having the timestamps correct is
really not a priority, compared to timely processing of the message
stream.  So let's use it whenever possible with QB_LOG_CONF_FILE_SYNC
requested.

[1] https://www.postgresql.org/message-id/1095055866.414539fadb90d@webmail.rawbw.com
    https://www.postgresql.org/message-id/attachment/20659/syncbench.c

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -201,8 +201,10 @@ AC_FUNC_CHOWN
 AC_FUNC_FORK
 AC_FUNC_MMAP
 AC_FUNC_STRERROR_R
-AC_CHECK_FUNCS([alarm clock_gettime ftruncate gettimeofday \
-		localtime localtime_r memset munmap socket \
+AC_CHECK_FUNCS([alarm clock_gettime \
+		fsync fdatasync ftruncate \
+		gettimeofday localtime localtime_r \
+		memset munmap socket \
 		strchr strrchr strdup strstr strcasecmp \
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -388,6 +391,31 @@ case "$host_os" in
...
 esac
 
+while : ; do
+	case ${file_sync} in
+	fdatasync)
+		if test x$HAVE_FDATASYNC != xyes; then
+			break
+		fi
+		file_sync=fsync;;
+	fsync)
```
jfriesse: I would suggest to add comment like "FALLTHROUGH is intentional"

jnpkrn: There were other issues (fixed not) but I'd say the iterated fallback
scheme is quite recognizable.

jfriesse: I would say it's not

jnpkrn: Ok, I'll add a brief explanation prior to start of the show.

Issue #265, opened 2017-07-31
Logging not working with binutils-2.29 standard linker (ld.bfd)


jnpkrn:
See http://oss.clusterlabs.org/pipermail/developers/2017-July/000504.html

chrissie-c:
Closing because pr#266 has been pulled.

PR #266, opened 2017-08-10
Workaround issues with ld from binutils 2.29

jnpkrn:
See #265.

~~WIP, do not merge yet, though feel free to comment.~~
EDIT: should no longer be considered WIP, now subject to review.


lib/Makefile.am
```diff
@@ -60,6 +66,29 @@ else
...
+	  while test -n "$${t1_bn}"; do t2_bn=$${t1_bn%.*[0-9]*}; \
+	  test "$${t2_bn}" != libqb.so || break; \
+	  test -L "$${t2_bn}" || { t1_bn=$${t2_bn}; continue; }; \
+	  t2_target=$$(ls -l "$${t2_bn}" || break); t2_target=$${t2_target#* -> }; \
+	  test "$${t2_target}" = "$${target}" || break; \
+	  t1_bn=$${t2_bn}; done; test -n "$${t1_bn}" || \
+	  { echo "only applicable to SO symlink scheme"; exit 1; }; \
+	  echo "INPUT($${t1_bn})" > "$(DESTDIR)$(libdir)/libqb.so-t"; \
+	  cat $< >> "$(DESTDIR)$(libdir)/libqb.so-t"; \
+	  mv -f "$(DESTDIR)$(libdir)/libqb.so-t" "$(DESTDIR)$(libdir)/libqb.so"
```
chrissie-c: This looks horrible. Can it be tidied up or put in a script somewhere?

jnpkrn: Perhaps (look at libtool itself to see some really hard to read shell code).

But what's worse, 12 lines of well-wrapped (and overall commented code),
or having to distribute yet another single-use-only file?  I am biased towards
the latter :)

jnpkrn:
Now passes "make distcheck".
Still rather WIP, but the most difficult part (making libtool actually serve
our purpose as a linker script dependency injector) is tackled now.

lib/Makefile.am
jnpkrn: Also, it gets a lot tidier when you observe `make install` screen :)

jnpkrn:
Ok, it now passes the CI run.  Another yet unexplored field is how this all combine
with _static_ linking, should anybody want to use that instead.

jnpkrn:
`WIP 3` had an issue of some unintended `RPATH` occurrences in
`qb-blackbox` binary or possibly `libqb.so.*`.

That's definitely not desired:
https://fedoraproject.org/wiki/Packaging:Guidelines#Beware_of_Rpath

Beside fixing that, `WIP 4` is also better commented (tries so).

lib/Makefile.am
jnpkrn: I was actually able to separate last 2 lines, leaving the "blob" down to 10 :)

jnpkrn:
Hopefully not many more iterations of this patchset is needed :)

jnpkrn:
> /usr/bin/ld: ../log_client.o: undefined reference to symbol 'dlclose@@GLIBC_2.2.5'

Strange, looks like `-ldl` disappeared from the build equations, will look into that.

jnpkrn:
OK, the issue is mainly that older libtool and gcc are not so smart
as the recent versions :-/
(libtool in not evaluating transitive accumulative closures, gcc for
being more picky about the order of arguments).
Plus some other negligence on my side.

jnpkrn:
Gotcha :-/
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=702737h

jnpkrn:
Testing on some other platforms, for freebsd, I need to at least respell the
sed command `lib/Makefile.am`.  Will see what's more to be observed outside
the Linux realms.

jnpkrn:
Fixed the sed command for portability and also bumped the `-version-info`
as necessitated by adding new API items -- I firmly think we should start
doing these bumps anytime the particular commit justifies/requires it.
Just consider the inter-release snapshots, they should all declare the
true status a not wait until some pre-release clarification, as otherwise
these snapshots could be flawed in terms of ABI compatibility, IMHO.

jnpkrn:
IOW, the reviewer should be obliged to spot "public ABI" changes and
ensure they are captured in `-version-info` properly.

kgaillot:
Development branches are not guaranteed to work in any shape or form, much less ABI compatibility. The reason ABI compatibility is not maintained in a development branch is so that the interface can change freely without breaking compatibility at release time. One commit may introduce ABI breakage, and a later commit restore it, so that at release time, client applications do not need to be recompiled. If the ABI version were maintained at every commit, it would necessitate a recompile.

jnpkrn:
> One commit may introduce ABI breakage, and a later commit restore it, so that at release time

Than the reviewers haven't done their homework. Figuring out the breakages just before the
release is way too late (to my taste).

Anyway, I need to harden the configure script check further as `ld.gold` (regardless of the
binutils version) will now fall into `attribute-section-workaround`category while later the
compilation will fail with:

```
error: lib/qblog_script.ld: SECTIONS seen after other input files; try -T/--script
```

Unlike with `ld.bfd`, where it's merely a warning and the linker can deal with it
(to our advantage).

jnpkrn:
And also when for instance compiling with native `cc` (a.k.a `clang`) in FreeBSD 11.1
(the newest release as of this comment), `/usr/bin/ld` is used for linking tasks, and
this is an ancient version: `GNU ld 2.17.50 [FreeBSD] 2007-07-03`.  But when one
uses GCC instead, it delegates the linking to `/usr/local/bin/ld` which is very recent
one: `GNU ld (GNU Binutils) 2.28`.  The point is, only the former chokes when
`qblog_script.ld` linker script is one of the input arguments so as to apply the
workaround when compiling libqb itself:
```
ld: section .hash [0000000000000158 -> 0000000000000b73] overlaps section __verbose [0000000000000000 -> 0000000000001837]
ld: .libs/libqb.so.0.19.0: section .hash lma 0x158 overlaps previous sections
ld: .libs/libqb.so.0.19.0: section .gnu.hash lma 0xb78 overlaps previous sections
ld: .libs/libqb.so.0.19.0: section .dynsym lma 0x1370 overlaps previous sections
ld: .libs/libqb.so.0.19.0: section .dynstr lma 0x3740 overlaps previous sections
ld: .libs/libqb.so.0.19.0: section .gnu.version lma 0x4ff4 overlaps previous sections
ld: .libs/libqb.so.0.19.0: section .gnu.version_r lma 0x52f0 overlaps previous sections
ld: .libs/libqb.so.0.19.0: section .rela.dyn lma 0x5370 overlaps previous sections
ld: .libs/libqb.so.0.19.0: section .rela.plt lma 0x8490 overlaps previous sections
ld: .libs/libqb.so.0.19.0: section .init lma 0x9c30 overlaps previous sections
ld: .libs/libqb.so.0.19.0: section .plt lma 0x9c44 overlaps previous sections
ld: .libs/libqb.so.0.19.0: section .text lma 0xac20 overlaps previous sections
ld: .libs/libqb.so.0.19.0: section .fini lma 0x1bfb8 overlaps previous sections
ld: .libs/libqb.so.0.19.0: section .rodata lma 0x1bfe0 overlaps previous sections
ld: .libs/libqb.so.0.19.0: section .eh_frame_hdr lma 0x1e758 overlaps previous sections
ld: .libs/libqb.so.0.19.0: section .eh_frame lma 0x1f2f8 overlaps previous sections
```
Just another reason to furthermore test, in the `configure` script the ability to apply
the workaround at all.

chrissie-c:
I'm with Ken on this one: ABI bumps should only happen on releases. Saying developers haven't "done their homework" is too unforgiving. People make mistakes, and ABIs evolve over development periods. People installing snapshots should be prepared for (even if we should do our best to minimise) instability.

jnpkrn:
@chrissie-c, but then, if reviewers missed that ABI changes, how could they
assess if it is justified and possibly offer a piece of advice to avoid it?
Seems pretty backwards, you don't want to deal with this just before the
release.  I see no conflict in praising reflecting ABI changes right away in
the commit to hit the `master` branch.  It's practical, it's self-describing,
it makes code easily shippable at any point.

chrissie-c:
We should do our best , of course. but mistakes happen. That aside, I just think that bumping ABI every time is too draconian. Releases are the development snapshots that are deemed to have been stabilised, THAT is when the version numbers are bumped up. Otherwise you're treating every minor commit as a release and committing to it. That's too much to expect.

Yes, I know we're taking about ABI changes and not every commit and they should be treated VERY carefully. but things happen, minds change, people make mistakes. Releases are what happens when we decide what things should be.

jnpkrn:
That's exactly why I am offering an automated ABI sanity check to support
my idea, exactly in order to eliminate human error. There are tools to figure
out such changes conveniently :)

chrissie-c:
s/mistakes/changes of mind/ ;-)

jnpkrn:
... that's addressed by another previous point: breaking ABI change should
go hand-in-hand with a new release (possibly joining several such deferred
changes).  Ironically enough, there was no such breaking change so far(!),
as otherwise we wouldn't be using `libqb.so.0` (or there was some cheating
already).  That will make anyone think twice whether the breakage is
justified, hopefully.

jnpkrn:
Added couple enhancements, most notably, I elaborated on the previously
mentioned idea of also libqb reverse-testing the client space, which is
now done by comparing known-to-be-of-libqb boundary symbols with
the one it globally sees, which should ideally be that of the client,
not a fall-through back to libqb.  In that case, libqb (as a proper library)
should not terminate the whole run, so it at least sends SOS to syslog :)

jnpkrn:
There's still one subtle detail: linker script enforces clients to carry section boundary denoting symbols whenever they compile with libqb -- even if the section is unused, which actually means a new conflict with the very original self-check via `QB_LOG_INIT_DATA`.  Another attempt to fix that failed.  I have one more idea to check, will see.

jnpkrn:
OK, this (really hopefully, it was quite a journey so far) last issue should be tackled now.
Help with testing the patchset is warmly welcome, in the (crafted or native) environment
(architecture, OS, compiler, linker), try for instance:
```
./autogen.sh \
&& ./configure \
&& make clean \
&& make -C lib \
&& make -C tests/functional/log_internal check
```

jnpkrn:
Indeed, there's one more point (ironic smile)...

When `QB_HAVE_ATTRIBUTE_SECTION` is not defined (happens on some arches),
the linker script seems still to be in place (should be dropped?) while the symbolic
names of the symbols are not evaluated because `<qb/qblog.h>` doesn't want to
know about these macro substitutions in that case.

jnpkrn:
Re previous comment:
Fortunately that was just a self-mystification of mine, and an example how
the problem is spreading in subtle ways (the described scenario was indeed
the case but the "badness" propagation happened a while ago when compiling
corosync with semi-fixed libqb, and now trying to compile pacemaker against
that + fix-complete libqb, the problem survived at those corosync libraries!).

So, disregard the previous self-FUDing and unless someone finds anything
wrong with the current patchset, I'd call it ready to merge in few days.

jnpkrn:
Based on feedback from @jfriesse, we should also do a better job of informing
users that the bad conditions detected in assertions of `QB_LOG_INIT_DATA`
could arise merely because of misusing this macro in a no-logging context.

jnpkrn:
Just gave a self-review another spin, merely tweaking commit or assertion
messages (see also https://github.com/ClusterLabs/libqb/pull/266#issuecomment-338804539).
For a consumable form of what has changed since the last revision, see
https://src.fedoraproject.org/rpms/libqb/c/e3be0a5f658dd3a81882b6a756efa8a880e1a26e?branch=master

jnpkrn:
Testing with Debian Mach+Hurd, I observe another issue with
`make -C tests/functional/log_internal check` arising from the fact it uses
completely shifted `errno` codes, i.e., `ENOENT` is 1073741826.

jnpkrn:
I shall tone down the priority of new self-check messages in `lib/log.c` as
this will also be triggered when someone purposefully compile the logging
client without callsite section. In that case, logging continues to be working
well.  And apparently the same file lacks `dlclose(work_handler)` at the
respective place,

jnpkrn:
Re previous comment:
- made both messages just a warning + previous error was extended with the info
  that perhaps just ` target chain built purposefully without callsite sections`
- added two `dlclose` calls that I think are not necessary at all as the underlying
  symbols spaces we in use before `dlopen` and remain to be in use, and the
  trigger for me to look at that -- valgrind warning -- was a spurious one that is not
  going away even after that change

And added a new commit introducing _a script to generate callsite-heavy logging client
so as to evaluate use of resources_.

jnpkrn:
On 07/11/17 09:33 +0000, Chrissie Caulfield wrote:
> A few small comments. The English in the error messages is a little
> clunky. In libqb.h, lines from L313 could be something more like:
> 
> "libqb's callsite section is observable, libqb's \
> +linkage is at fault and logging may not work reliably"
> 
> "libqb's callsite section is non-empty, libqb's \
> +linkage is at fault and logging may not work reliably"		
> 
> And similar at lines 328, 333, 376, 384. If you're not sure then ask
> me on IRC :)

Thanks, will take a look.

> Also, I like the qb_version struct but I don't think it's relevant
> to this PR so I think it should be removed and submitted separately
> to keep the commit logs tidy.

It's related in a way that it would be too easy to forget to fix the
reliance of the newly introduced self-checks (introduced/extended in
the same PR) on the symbol that now gets turned static/private
(because it was visible merely by accident before) in favour of those
new version-carrying symbols (explicitly public) -- I prefer to stay
safe and, with refreshed memory, piggy-back that commit to that PR
presenting a logical unit, which aims to reconcile visibility of
(other) symbols anyway.  And if anyone would do a mere backport, I'd
prefer the PR was treated as an atomic change (modulo logging client
generating script perhaps, as I do for Fedora Rawhide), which would
be less apparent with a totally separate commit.

Will this justification stand the judgement?

-- 
Poki

chrissie-c:
OK, I'll let the version struct stay here. At this stage I suspect we're both weary from this whole process and want to get it over with ;-)

jnpkrn:
On 07/11/17 09:33 +0000, Chrissie Caulfield wrote:
> A few small comments. The English in the error messages is a little
> clunky. In libqb.h, lines from L313 could be something more like:
> 
> "libqb's callsite section is observable, libqb's \
> +linkage is at fault and logging may not work reliably"

Context:
Original did not contain "is" and used "would not" instead of "may not".

Let me explain:

- it's a string expression in the assert statement that normally is
  not there (one is informed solely about the failed condition), but
  I felt much better diagnostics is sort of an obligation here, still
  the news-headline style brevity should be OK here

- "may not" would be fine, but the program already got terminated at
  that point, so let's signal the precondition for the reasonable
  function of the program hasn't been fulfilled and for that it seems
  saying "something would not work (anyway)" an appropriate expression

- those assertions and messages are not expected to be hit after the
  fix in full, but are kept around just in case.

What do you think about those messages in this light?

Also, I was toying with the idea of adding environment variable
override to suppress those checks by force, but left it behind, as
really it should be exceptional case that those assertions will hit
with the fix, and the easiest solution is recompiling the client
program with either or both of -DQB_KILL_ATTRIBUTE_SECTION and
-DNDEBUG (and for, $deity forgive, programs without a source, there
should always be a possibility to hijack __assert_fail or equivalent
with an LD_PRELOAD trick as a last resort).


But ... I took a stab at not introducing any new roadblocks for having
libqb cross-buildable (see
https://github.com/ClusterLabs/libqb/pull/269#issuecomment-339770888
and on) and it looks auspiciously.

-- 
Poki

jnpkrn:
On 14/11/17 20:47 +0100, Jan Pokorný wrote:
> Context:
> Original did not contain "is" and used "would not" instead of "may not".
> 
> [...]
> 
> - "may not" would be fine, but the program already got terminated at
>   that point, so let's signal the precondition for the reasonable
>   function of the program hasn't been fulfilled and for that it seems
>   saying "something would not work (anyway)" an appropriate expression

And in contrast, in log.c, the warnings are just logged and the flow
of qb_log_init and on continues as usual, and for that "will not"
wording is used -- apparently something is broken (except for
expressly given conditions) so stick with a stringent language.

Does it make sense?

-- 
Poki

jnpkrn:
Made the configure script avoid a run-time check for compatibility with
cross-building as discussed.  That catches all binutils 2.28, 2.29, 2.29.1
cases as it should in my testing.

Plus added an explicit not on the topic of spurious ld.bfd warning after
the fix:

```
   NOTE: When concerned about a warning coming from the build process like
         warning: [...]libqb.so contains output sections; did you forget -T?
         while it finishes OK, consider it merely a harmless side-effect.
```

in `libqb.pc` and `libqb.so` (now a linker script) files.

chrissie-c:
I'm  not quibbling about the logic in the messages, I'm concerned about the English. To a native speaker they are almost incomprehensible - which is probably why I misunderstood them ;-)

jnpkrn:
Chrissie, you can spot the gist of wording changes at
https://src.fedoraproject.org/rpms/libqb/c/93fffaf298477d0d4c4a0a6898a501d6f068a43a

I also demoted one of the new non-fatal warnings in `lib/log.c` to mere notice.

chrissie-c:
Now merged. Thanks again for all your work on this issue

jnpkrn:
Some after the fact notes:
* @jfriesse noted FreePascal actually overcome the "did you forget -T" warning,
  for which I gathered these references:
  - https://svn.freepascal.org/cgi-bin/viewvc.cgi?view=revision&revision=31459
  - https://svn.freepascal.org/cgi-bin/viewvc.cgi?view=revision&revision=31460
  - https://svn.freepascal.org/cgi-bin/viewvc.cgi?view=revision&revision=31675
  - all via https://bugs.freepascal.org/view.php?id=29228
* nevertheless, Gold linker still doesn't allow for `INSERT` command in the
  linker script (now meaning properly passed `-T <script>` one):
  - https://sourceware.org/bugzilla/show_bug.cgi?id=15373

  so the only imaginable gain to do anything more now is to get rid of said
  warning, at the cost of:
  - level of overall hackery increased even more (per the referred FreePascal
    changes)
  - dropping any sort of support for plain `-lqb` in build flags
    (effectively enforcing using `pkg-config` tool to figure the flags out):

gao-yan:
```
And then for 2.29.1 release of binutils once again, as someone actually
noticed something went overboard with the 2.29 changes:

http://lists.gnu.org/archive/html/bug-binutils/2017-08/msg00195.html
...
https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;h=487b6440dad57440939fab7afdd84a218b612796

At least that change doesn't invalidate all the effort being put into
the original version of the changeset, only the configure script check
had to be refined so as not to miss the "orphan section magic not
working properly out of the box, without band aid" observation
(see the inline comment) -- the workaround arrangement needs
to be applied in that case as well.
```
Does the fix in binutils-2.29.1 provide any practical help for this issue? It sounds like basically libqb has to do the exactly same for binutils-2.29.1 as for binutils-2.29?

gao-yan:
And there seems to be some issue with this combination:

1) libqb-1.0.3+20171226.6d62b64 built with binutils-2.26.1:
checking whether GCC supports __attribute__((section()) + ld supports orphan sections... yes
checking whether linker emits global boundary symbols for orphan sections... yes

2) And pacemaker-1.1.18+20180104.7ba28d854 fails to build with this libqb:
```
PATH=/home/abuild/rpmbuild/BUILD/pacemaker-1.1.18+20180104.7ba28d854/tools:$PATH /home/abuild/rpmbuild/BUILD/pacemaker-1.1.18+20180104.7ba28d854/tools/iso8601 --help
...
iso8601: utils.c:66: common: Assertion `"implicit callsite section is populated, otherwise target's build is at fault, preventing reliable logging" && __start___verbose != __stop___verbose' failed.
Makefile:1469: recipe for target 'iso8601.8' failed
gmake[1]: *** [iso8601.8] Aborted (core dumped)
```
-- This is on openSUSE 42.3. 

With the same binutils and other package versions on SLE12 SP3/SP2,  "crm_shadow --help" triggers the same assertion on build.

jnpkrn:
On 11/01/18 07:29 +0000, Gao,Yan wrote:
> Does the fix in binutils-2.29.1 provide any practical help for this
> issue? It sounds like basically libqb has to do the exactly same for
> binutils-2.29.1 as for binutils-2.29?

I was also interested back when 2.29.1 arrived as I was curious
whether the time invested to deal with 2.29 was worth it would it be
just for the only linker versions, but it turned up that nope,
private->protected shift in symbols visibility between 2.29 and 2.29.1
is not enough, as original libqb counted on public one (it's my
understanding only this models situation with static linking the best,
though never tested, because such practice is dedicated to very
special cases nowadays).

And yes, part of the point is that for "build-time link with libqb",
there needs to be rather unified handling for whatever (supported)
linker will be used (otherwise LDFLAGS would need to look like
$(qbconfig --libs) for the dependent components, where qbconfig would
internally guestimate [unreliably!] which linker will be used and
arranged the actual flags accordingly, which looks pretty cumbersome),
which is exactly the current state -- the overlay linker script will
not hurt linker from binutils < 2.29 either, and keeps log module
working well with >= 2.29.

As I mentioned in the recent Fedora conversation:
https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/message/UIACQTZZYIS627AUREOOOPU5MFGC3HSW/
I was originally unaware about the separate upstream discussion,
outcome of which was that I had to be more strict in configure checks
to capture 2.29.1 in addition to 2.29 where the former iteration of
them work well.

Hope this helps.

-- 
Poki

jnpkrn:
On 11/01/18 07:58 +0000, Gao,Yan wrote:
> And there seems to be some issue with this combination:
> 
> 1) libqb-1.0.3+20171226.6d62b64 built with binutils-2.26.1:
> checking whether GCC supports __attribute__((section()) + ld supports orphan sections... yes
> checking whether linker emits global boundary symbols for orphan sections... yes
> 
> 2) And pacemaker-1.1.18+20180104.7ba28d854 fails to build with this libqb:
> ```
> PATH=/home/abuild/rpmbuild/BUILD/pacemaker-1.1.18+20180104.7ba28d854/tools:$PATH /home/abuild/rpmbuild/BUILD/pacemaker-1.1.18+20180104.7ba28d854/tools/iso8601 --help
> ...
> iso8601: utils.c:66: common: Assertion `"implicit callsite section is populated, otherwise target's build is at fault, preventing reliable logging" && __start___verbose != __stop___verbose' failed.
> Makefile:1469: recipe for target 'iso8601.8' failed
> gmake[1]: *** [iso8601.8] Aborted (core dumped)
> ```
> -- This is on openSUSE 42.3. 
> 
> With the same binutils and other package versions on SLE12 SP3/SP2,  "crm_shadow --help" triggers the same assertion on build. 

I've observed various issues in the past including something like
this (failures when building pacemaker), let me mimic your combination
and I'll get back to you.

-- 
Poki

jnpkrn:
On 11/01/18 09:55 +0100, Jan Pokorný wrote:
> On 11/01/18 07:58 +0000, Gao,Yan wrote:
>> And there seems to be some issue with this combination:
>> 
>> 1) libqb-1.0.3+20171226.6d62b64 built with binutils-2.26.1:
>> checking whether GCC supports __attribute__((section()) + ld supports orphan sections... yes
>> checking whether linker emits global boundary symbols for orphan sections... yes
>> 
>> 2) And pacemaker-1.1.18+20180104.7ba28d854 fails to build with this libqb:
>> ```
>> PATH=/home/abuild/rpmbuild/BUILD/pacemaker-1.1.18+20180104.7ba28d854/tools:$PATH /home/abuild/rpmbuild/BUILD/pacemaker-1.1.18+20180104.7ba28d854/tools/iso8601 --help
>> ...
>> iso8601: utils.c:66: common: Assertion `"implicit callsite section is populated, otherwise target's build is at fault, preventing reliable logging" && __start___verbose != __stop___verbose' failed.
>> Makefile:1469: recipe for target 'iso8601.8' failed
>> gmake[1]: *** [iso8601.8] Aborted (core dumped)
>> ```
>> -- This is on openSUSE 42.3. 
>> 
>> With the same binutils and other package versions on SLE12 SP3/SP2,  "crm_shadow --help" triggers the same assertion on build. 
> 
> I've observed various issues in the past including something like
> this (failures when building pacemaker), let me mimic your combination
> and I'll get back to you.

Reproduced with (and with both 2.28 and 2.29, actually):
./log_test_mock.sh -ncl -tqb+_il+_c+ ../../libqb-1.0.3-1.1.6d62.*.src.rpm

(see also https://github.com/ClusterLabs/libqb/pull/288 where I hope
to eventually develop a solution to this fallout).

-- 
Jan (Poki)

jnpkrn:
On 11/01/18 21:12 +0100, Jan Pokorný wrote:
> On 11/01/18 09:55 +0100, Jan Pokorný wrote:
>> On 11/01/18 07:58 +0000, Gao,Yan wrote:
>>> And there seems to be some issue with this combination:
>>> 
>>> 1) libqb-1.0.3+20171226.6d62b64 built with binutils-2.26.1:
>>> checking whether GCC supports __attribute__((section()) + ld supports orphan sections... yes
>>> checking whether linker emits global boundary symbols for orphan sections... yes
>>> 
>>> 2) And pacemaker-1.1.18+20180104.7ba28d854 fails to build with this libqb:
>>> ```
>>> PATH=/home/abuild/rpmbuild/BUILD/pacemaker-1.1.18+20180104.7ba28d854/tools:$PATH /home/abuild/rpmbuild/BUILD/pacemaker-1.1.18+20180104.7ba28d854/tools/iso8601 --help
>>> ...
>>> iso8601: utils.c:66: common: Assertion `"implicit callsite section is populated, otherwise target's build is at fault, preventing reliable logging" && __start___verbose != __stop___verbose' failed.
>>> Makefile:1469: recipe for target 'iso8601.8' failed
>>> gmake[1]: *** [iso8601.8] Aborted (core dumped)
>>> ```
>>> -- This is on openSUSE 42.3. 
>>> 
>>> With the same binutils and other package versions on SLE12 SP3/SP2,  "crm_shadow --help" triggers the same assertion on build. 
>> 
>> I've observed various issues in the past including something like
>> this (failures when building pacemaker), let me mimic your combination
>> and I'll get back to you.
> 
> Reproduced with (and with both 2.28 and 2.29, actually):
> ./log_test_mock.sh -ncl -tqb+_il+_c+ ../../libqb-1.0.3-1.1.6d62.*.src.rpm
> 
> (see also https://github.com/ClusterLabs/libqb/pull/288 where I hope
> to eventually develop a solution to this fallout).

Ok, I figured why I haven't observed any issue like that e.g. when
building pacemaker just like you, except I did it for Fedora.
As mentioned, I observe the problem manifested when mimicking the same
with "./log_test_mock.sh -ncl" test runner, unless the advice bellow
is applied.

The crucial point Fedora does is that it injects "-Wl,-z,relro" to
%__global_ldflags RPM macro which is then used in %configure one,
which is in turn used in pacemaker.spec.

So as an interim solution, try a line like this at the top of
pacemaker.spec (untested, just that you grok the idea):

  %define __global_ldflags %{expand:%__global_ldflags} -Wl,-z,relro

Build process should then pass OK and the resulting binaries should
work without an issue.

-- 
Jan (Poki)

jnpkrn:
Anyway, another solution would be fix the terrifying mess representing
current shape of pacemaker's linking declarations, another technical
debt impeding the code base sanity.  For instance, it's a nonsense to
link test.iso8601.o with libqb directly, to begin with.

Note that to avoid a transitive transfer of libraries to link with
from immediate libtool libraries (*.la), which would apparently be
needed so as to avoid CLI tools that don't use libqb directly not
be linked with it like that (good), one would also need to define
"link_all_deplibs=no" in configure script (this is also Debian/Ubuntu
default).

However, it would not solve the problem completely, as there may be
cases, e.g., amongst daemons, where libqb is used directly but no
logging message is arranged in the immediate code of that program.

Hence, using above link flag as an interim solution is more appealing,
till some more systemic tweak is discovered.

-- 
Jan (Poki)

gao-yan:
FYI, there's a thread about this topic ongoing from binutils upstream:
https://sourceware.org/ml/binutils/2018-01/msg00265.html

jnpkrn:
Thanks for the heads up.
I am afraid I'm not qualified to contribute to the discussion about these
internal details, I am barely knowledgable about the field from the user's
perspective.  One of the biggest barriers to entry is the non-existing
documentation - it's just the code that deliberately decides wha't the
right way to go.

It looks to me that current approach, together with some rough edges
removal (#288), works well for us.  Yes, there's a crux of superfluous
warning being emitted by the linker when workaround linker script
kicks in, but that was voiced just once and the explanation is handy
at places one might consequently look at.

If there's any intention to handle the mechanism at hand better on linker
side, I think it's natural to strictly require that our workaround approach
keeps working regardless.  AFAIK, this already applies to all reasonably
recent binutils versions, pre-2.29 - 2.30, and would be very pitiful if the
"tweak for good" actually broke this precious continuity...

gao-yan:
I can tell the difficulty of it... You did nice work fixing this though.

Apparently binutils upstream realized the situation indeed needs to be improved and they are  achieving a better solution now:

https://sourceware.org/ml/binutils/2018-02/msg00056.html

Actually more users are  aware of that:
https://sourceware.org/ml/binutils/2018-02/msg00038.html

I also hope the resulting solution won't break the current approach of libqb.

jnpkrn:
Looking at

https://sourceware.org/ml/binutils/2018-01/msg00347.html

and figuring out that there have been more changes on the topic recently
in master:
```
$ git log --oneline --since=2017-11-01 | grep -Ee __start -e __stop
823143c6ca Check if __start/__stop symbols are referenced by shared objects
040b4a9eb8 Add --gc-sections test checking removal of __start/__stop symbols.
36b8fda5d6 Make __start/__stop symbols dynamic and add testcase
32253bb796 Define __start/__stop symbols when there is only a dynamic def
```
I got slightly worried.

Good news is that libqb's internal sanity check:
```
./autogen.sh && ./configure && make clean && make -C lib \
  && make -C tests/functional/log_internal/ check
```
keeps working also with this built-from-checkout binutils version,
407aa07cee - 2.30.0.20180207
(binutils compilation was choking on some incompatible function
signatures with GCC 8 so had to resort to downgrade, sigh).

wferi:
@jnpkrn, while uploading Corosync 2.4.4 (built with libqb 1.0.3) to Debian unstable, I noticed that the size of the `corosync` binary grew from 314632 bytes (in 2.4.2, built with libqb 1.0.1) to 1048376 bytes. According to `readelf -l` the sizes of the ELF segments didn't change much, but the `__verbose` section moved into a new loadable segment with file offset 0xfa820. Such a large offset seems unreasonable, the file is pure zeros between 0x48600 and 0xfa820 and no segment uses that range. Can you perhaps offer any insight?
Thanks,
Feri.

gao-yan:
BTW, of course we already have this linker script to keep these symbols "GLOBAL DEFAULT" , which is relied on by the current code I believe. Otherwise the resulting "GLOBAL PROTECTED" symbols generated by the new binutils won't work.

My colleague, Michael Matz, who works on binutils reviewed the relevant code of libqb, and is pointing out:
```
No idea, why this should have worked in the past but I think this is a genuine
bug in the init function of the logger.

Basically it wants to go through all loaded shared objects, looks for
__start___verbose/__stop___verbose symbols in each of them and then initializes
the descriptors between both addresses.  The symbol lookup is done via dlsym.

Now, that is all fine and dandy in principle, but the callback that collects
all loaded objects (via dl_iterate_phdr) is:

static int32_t
_log_so_walk_callback(struct dl_phdr_info *info, size_t size, void *data)
{
        struct dlname *dlname;

        if (strlen(info->dlpi_name) > 0) {
                dlname = calloc(1, sizeof(struct dlname));
                if (!dlname)
                        return 0;
                dlname->dln_name = strdup(info->dlpi_name);
                qb_list_add_tail(&dlname->list, &dlnames);
        }

        return 0;
}

(in libqb/lib/log.c).  Not the check for strlen(info->dlpi_name) > 0.
That ignores the shared object created for the executable itself (it always
has a empty name).  If also find not other code which would try to enumerate
the __verbose section from the executables itself.

That matches the effects of the testsuite: some log messages (from libqb itself
and some other shared libs) are emitted just fine.  In addition the messages from crm_debug and crm_trace are also registered because they use a different
mechanism (not via the __verbose sections).  But the normal log messages
from the executable (e.g. from pacemaker/fencing/commands.c:926) are missing
as they aren't registered.

What needs to happen in the log initializer is that also the symbols of the main
object are handled.  For this you'd normally use "dlopen(NULL, flags)".
I bet extending qb_log_fini (or it's subroutine _log_so_walk_dlnames) to not
forget the main object will make everything work.
```
Could this be an approach that we'd still like to investigate?

gao-yan:
Nice Michael Matz <matz@suse.de> kindly offered this patch, which passes the tests with or without this pull request https://github.com/ClusterLabs/libqb/pull/266 by using binutils with the patches back-ported from binutils-2.30:

```
--- libqb-1.0.1.22-59ea.orig/lib/log.c	2017-02-15 17:32:45.000000000 +0100
+++ libqb-1.0.1.22-59ea.orig/lib/log.c	2018-04-20 15:43:02.000000000 +0200
@@ -781,44 +781,50 @@ _log_so_walk_callback(struct dl_phdr_inf
 }
 
 static void
+_log_register_one(const char *dlname)
+{
+	void *handle, *start, *stop;
+	const char *error;
+
+	handle = dlopen(dlname, RTLD_LAZY);
+	error = dlerror();
+	if (!handle || error) {
+		qb_log(LOG_ERR, "%s", error);
+		goto done;
+	}
+
+	start = dlsym(handle, QB_ATTR_SECTION_START_STR);
+	error = dlerror();
+	if (error) {
+		goto done;
+	}
+
+	stop = dlsym(handle, QB_ATTR_SECTION_STOP_STR);
+	error = dlerror();
+	if (error) {
+		goto done;
+
+	} else {
+		qb_log_callsites_register(start, stop);
+	}
+done:
+	if (handle)
+		dlclose(handle);
+}
+
+static void
 _log_so_walk_dlnames(void)
 {
 	struct dlname *dlname;
 	struct qb_list_head *iter;
 	struct qb_list_head *next;
 
-	void *handle;
-	void *start;
-	void *stop;
-	const char *error;
-
+	_log_register_one(NULL);
 	qb_list_for_each_safe(iter, next, &dlnames) {
 		dlname = qb_list_entry(iter, struct dlname, list);
 
-		handle = dlopen(dlname->dln_name, RTLD_LAZY);
-		error = dlerror();
-		if (!handle || error) {
-			qb_log(LOG_ERR, "%s", error);
-			goto done;
-		}
-
-		start = dlsym(handle, QB_ATTR_SECTION_START_STR);
-		error = dlerror();
-		if (error) {
-			goto done;
-		}
-
-		stop = dlsym(handle, QB_ATTR_SECTION_STOP_STR);
-		error = dlerror();
-		if (error) {
-			goto done;
-
-		} else {
-			qb_log_callsites_register(start, stop);
-		}
-done:
-		if (handle)
-			dlclose(handle);
+		_log_register_one(dlname->dln_name);
+
 		qb_list_del(iter);
 		if (dlname->dln_name)
 			free(dlname->dln_name);
```

gao-yan:
The revisions of binutils that matter:

commit 8dfb7cbf8401be97077f5919ac7473bdbfa8b692
Author: H.J. Lu <hjl.tools@gmail.com>
Date:   Tue Aug 22 09:41:21 2017 -0700

    Update PR ld/21964 tests

commit 32253bb7963ac7caa166ec41e336372f2ffc03d4
Author: Alan Modra <amodra@gmail.com>
Date:   Tue Jan 23 10:50:02 2018 +1030

    Define __start/__stop symbols when there is only a dynamic def

commit 36b8fda5d614cb5aaf701a92befa9919bd0b195a
Author: Alan Modra <amodra@gmail.com>
Date:   Mon Jan 29 21:45:09 2018 +1030

    Make __start/__stop symbols dynamic and add testcase

commit 823143c6ca8ef4267e67ba03771991e08d09fabd
Author: H.J. Lu <hjl.tools@gmail.com>
Date:   Wed Jan 31 05:10:40 2018 -0800

    Check if __start/__stop symbols are referenced by shared objects

commit bf3077a6c3c9ff21c072a6f42c91bffefd35bc15
Author: Michael Matz <matz@suse.de>
Date:   Wed Jan 31 14:26:46 2018 +0100

    bfd_elf_define_start_stop: Fix check

gao-yan:
BTW, this is the comment that Michael Matz made when proposing the patch:
```
My theory why it worked before
at all is that the __start___verbose symbols were not PROTECTED but DEFAULT (i.e. global), so looking up that symbol for a shared lib that didn't contain
those symbols (there are many) actually found the syms from the main object.
Now that they are PROTECTED that doesn't happen anymore.  So the old
binutils behaviour hid this bug, but now doesn't anymore.
```

wferi:
@jnpkrn, I've got no Fedora VMs handy. Furthermore, I haven't done any testing yet, just noticed the changed section layout which made me think you might have relevant info to share before I start rabbit-holing that fatty binary... I'm not even sure libqb has anything to do with this, it's just a guess at the moment. But I'll start collecting hard data eventually.

gao-yan:
Oh, this reminds me that Machael Matz once mentioned:
```
I'll also note that the trick you're using to work-around the binutils problem
(i.e. the linker script enforcing a __verbose section with boundary symbols) has
the side-effect of allocating a new segment for that section, which due to
alignment reasons is placed very far away from the existing ones.  That's not
a problem in itself, it just costs a bit memory, but I though to mention it
anyway.  It might be possible to play with the linker script to fold that
__verbose section into the normal data segment.
```

wferi:
@gao-yan, that must be it, thanks for sharing! Michael Matz obviously knows his stuff.

PR #271, opened 2017-10-12
WIP: Hurd portability

wferi:


chrissie-c:
This failed CI on the monotonic check. Can you fix please?

wferi:
Yes, I know, this only works on Hurd at the moment (thus the WIP tag). Looks like it isn't possible to check for a working monotonic clock at configuration or even compile time, but a runtime check will be needed, because for example on x86 `_POSIX_MONOTONIC_CLOCK` is defined to 0 (ask `sysconf`). Also, on Hurd `CLOCK_MONOTONIC` is actually `CLOCK_REALTIME` behind the scenes. I'm not sure how much effort this mess warrants at the end of the day.

wferi:
The monotonic stuff is more involved and mostly independent, let's separate it out.

wferi:
Wait, something funny is still going on with the permissions...

wferi:
Nevermind, it was a chroot artifact.

PR #277, opened 2017-11-15
maint: make -devel package dependency on the main package arch-qualified

jnpkrn:
Beside the fact that libqb.pc is indeed arch-specific (see "libdir"
variable), it's also a matter of being disciplined per what distros
require:
https://fedoraproject.org/wiki/Packaging:Guidelines#Requiring_Base_Package

The syntactic provision at hand was introduced in rpm 4.6.0
(8+ years ago): http://rpm.org/user_doc/arch_dependencies.html


chrissie-c:
Merged. thanks.

PR #279, opened 2017-12-13
Doc clarifications

jnpkrn:


include/qb/qblog.h
```diff
@@ -113,13 +118,18 @@ extern "C" {
  *	qb_log_ctl(mytarget, QB_LOG_CONF_THREADED, QB_TRUE);
  * @endcode
  *
- * To workaround your syslog daemon filtering all messages > LOG_INFO
+ * To workaround your syslog daemon filtering messages with the priority of
+ * @c LOG_DEBUG or of lesser importance while such messages should still get
+ * across without any system-wide changes (here, proportionally tweaking the
+ * priority of these such that @c LOG_DEBUG gets promoted to @c LOG_INFO):
  * @code
```
chrissie-c: I think the original message is fine here. The comment here is too wordy and complex and reads very badly in English.

include/qb/qblog.h
```diff
@@ -113,13 +118,18 @@ extern "C" {
...
  *	qb_log_ctl(QB_LOG_SYSLOG, QB_LOG_CONF_PRIORITY_BUMP,
  *		   LOG_INFO - LOG_DEBUG);
  * @endcode
  *
- * To ensure all logs to file targets are fsync'ed (default QB_FALSE)
+ * To ensure all logs to file targets are fsync'ed (new messages expressly
+ * run aground at the storage level as they keep coming, otherwise defaults
+ * to @c QB_FALSE):
  * @code
```
chrissie-c: 'run aground' sounds like it's been translated into Pirate :) Flushed to storage would be better, though I'm pretty sure people reading this will understand what fsync does anyway. Rephrase this please or(as above) just leave it as-is.

include/qb/qblog.h
jnpkrn: So it was just me, who wasn't able to decipher what exactly to expect
from using that configuration toggle?  I'd doubt that.  Are you able
to tell from the original description, without looking into source,
and forgetting my rewording?

include/qb/qblog.h
jnpkrn: Yes, honestly that term was found in the dictionary, as I was
intentionally after something rather vague, because with all this
smartness on all the levels of storage stack, very few guarantees
about the state of data can be given.  And "flushed" is overloaded
with `flush`, which is something different.

include/qb/qblog.h
chrissie-c: IT seems pretty clear what happens - even if WHY it happens is a total mystery! If you can come up with something clearer I'll happily include it. But I'm afraid that's incomprehensible English

PR #281, opened 2017-12-19
Fix some warnings (mostly coming from newly enabled build switches) + standard maintenance scrutiny

jnpkrn:


jnpkrn:
Hmm, some resemblances with real issues:
https://bugzilla.redhat.com/show_bug.cgi?id=1475482#c16

jnpkrn:
Non-isulated `-Wsign-compare` cases are quite tough to decide responsibly,
so if it's preferred to have a release sooner rather than later, @chrissie-c,
please accept what's here so far.

chrissie-c:
Thanks for all those! I've added a small fix to the warnings that only appear on RHEL7 too.

PR #285, opened 2017-12-25
build: fix configure script neglecting, re-enable out-of-tree builds

jnpkrn:
For the former, a prototype and the final code got (hm, mysteriously)
intertwisted.  For the latter, I am clearly guilty of (rare, anyway)
testing of the out-of-tree builds only with libqb-already-system-wide
scenario, which is rather shortsighted.

Thanks Fabio and his ci.kronosnet.org project for spotting that.


jnpkrn:
Actually, we are in the situation that lack of this patch blocks knet CI testing,
so daring to make Xmas exception in the review process and push right away.
(That's by no means a standard process.)

fabbione:
Thanks for the quick fix, and it´s working fine afaict.

PR #288, opened 2018-01-11
Binutils 2.29 followup

jnpkrn:
see also https://github.com/ClusterLabs/libqb/pull/266#issuecomment-356855212


jnpkrn:
Current patchset (ending with 03497c7) looks good to me so far,
but needs some more testing and proper commit message for the
last commit.  Will work some more but would welcome some feedback.
@gao-yan, any?

Regarding testing, I realized that protected visibility of the section boundary
symbols (either enforced or native with binutils 2.29.1, btw. making 2.29
technically the most problematic binutils release from the libqb's perspective)
can come handy for SW not using `_GNU_SOURCE` a.k.a. GNU extensions
on top of POSIX defined interfaces so I arranged the originally for-all check just
for "POSIX only" cases together with sais visibility enforcement, unless overridden
with `QB_LD_2_29` because thats the only show stopper for that and we cannot
really figure out which linker is being used declaratively.

gao-yan:
Thanks for all the effort on this!

Tried libqb-1.0.3+20171226.6d62b64 plus https://github.com/ClusterLabs/libqb/pull/288/commits/03497c7bde663821b63253aa09ff788a6809b04  with binutils-2.26.1.

pacemaker-1.1.18+20180112.bd7115b8c compiles now :)

```
# readelf -s /usr/lib64/libqb.so.0.19.0 | grep verbose
   162: 0000000000228058     0 NOTYPE  GLOBAL DEFAULT   26 __stop___verbose
   239: 0000000000226870     0 NOTYPE  GLOBAL DEFAULT   26 __start___verbose
# readelf -s /usr/lib/pacemaker/stonithd | grep verbose
   163: 0000000000a1ddd0     0 NOTYPE  GLOBAL DEFAULT bad section index[ 36] __start___verbose
   166: 0000000000a1f270     0 NOTYPE  GLOBAL DEFAULT bad section index[ 36] __stop___verbose
```
--- No idea what it's supposed to mean ;) I think I've seen the similar with the things linked with binutils-2.29.1 too.

Fencing regression tests (which reply on logging) all passed.

Pacemaker seems to work well, but every time on shutting down cluster services, crmd encounters:
```
Jan 18 15:17:46 f171 crmd[30523]:     crit: GLib: Source ID 22 was not found when attempting to remove it
```

With the same sources but linked with binutils-2.29.1, it looks similar and has an additional error from crm_glib_handler:

```
Jan 18 15:41:13 f58.suse.de crmd[10048]:    error: crm_glib_handler: Forked child 10080 to record non-fatal assert at logging.c:73 : Source ID 23 was not found when attempting to remove it
Jan 18 15:41:13 f58.suse.de crmd[10048]:     crit: GLib: Source ID 23 was not found when attempting to remove it
```

The relevant code in pacemaker/lib/common/logging.c: crm_glib_handler()
```
  63         glib_cs = qb_log_callsite_get(__FUNCTION__, __FILE__, "glib-handler", LOG_DEBUG, __LINE__, c│test-driver
     rm_trace_nonlog);                                                                                   │test.patch
  64     }                                                                                               │test-simple6.xml
  65                                                                                                     │test-unmanaged-block-restart.xml
  66                                                                                                     │test-unmanaged-stop-4.xml
  67     switch (msg_level) {                                                                            │test.xml
  68         case G_LOG_LEVEL_CRITICAL:                                                                  │ticket-site-attr.xml.orig
  69             log_level = LOG_CRIT;                                                                   │tools
  70                                                                                                     │travisci_build_coverity_scan.sh
  71             if (crm_is_callsite_active(glib_cs, LOG_DEBUG, 0) == FALSE) {                           │unmanaged-block-restart-try.patch
  72                 /* log and record how we got here */                                                │unmanaged-dep.xml
  73                 crm_abort(__FILE__, __FUNCTION__, __LINE__, message, TRUE, TRUE);                   │unmanaged-stop-5
  74             }                                                                                       │unmanaged-stop-5.xml
  75             break;                                                                                  │valgrind-pcmk.suppressions
  76                                  
```
Not sure if it's necessarily related this change.

kgaillot:
> Pacemaker seems to work well, but every time on shutting down cluster services, crmd encounters:
> 
> Jan 18 15:17:46 f171 crmd[30523]:     crit: GLib: Source ID 22 was not found when attempting to remove it
> 
> With the same sources but linked with binutils-2.29.1, crmd encounters the similar on shutting down and has an additional error from crm_glib_handler:
> 
> Jan 18 15:41:13 f58.suse.de crmd[10048]:    error: crm_glib_handler: Forked child 10080 to record non-fatal assert at logging.c:73 : Source ID 23 was not found when attempting to remove it
> Jan 18 15:41:13 f58.suse.de crmd[10048]:     crit: GLib: Source ID 23 was not found when attempting to remove it

No idea what's going on there. I'm particularly concerned that with 2.26.1 crm_glib_handler() appears not to be called for a glib error. I wonder if a similar linker issue is affecting glib itself.

Most likely the particular glib error involves calling g_source_remove() on a timer. Enabling trace logs would probably narrow down which one and where. Since it's at shutdown, maybe it's the throttle timer.

It might be useful to try with different glib versions and/or glib+linker combinations to see if some behavior changed there rather than in the linker, libqb, or pacemaker. But that's more work than I want to put into this :)

gao-yan:
Well, without https://github.com/ClusterLabs/libqb/commit/03497c7bde663821b63253aa09ff788a6809b04, crmd encounters the same.

jnpkrn:
On 18/01/18 15:06 +0000, Gao,Yan wrote:
> Thanks for all the effort on this!
> 
> Tried libqb-1.0.3+20171226.6d62b64 plus
> https://github.com/ClusterLabs/libqb/pull/288/commits/03497c7bde663821b63253aa09ff788a6809b04
> with binutils-2.26.1.
> 
> pacemaker-1.1.18+20180112.bd7115b8c compiles now :)

Perfect!

> ```
> # readelf -s /usr/lib64/libqb.so.0.19.0 | grep verbose
>    162: 0000000000228058     0 NOTYPE  GLOBAL DEFAULT   26 __stop___verbose
>    239: 0000000000226870     0 NOTYPE  GLOBAL DEFAULT   26 __start___verbose
> # readelf -s /usr/lib/pacemaker/stonithd | grep verbose
>    163: 0000000000a1ddd0     0 NOTYPE  GLOBAL DEFAULT bad section index[ 36] __start___verbose
>    166: 0000000000a1f270     0 NOTYPE  GLOBAL DEFAULT bad section index[ 36] __stop___verbose
> ```
> --- No idea what it's supposed to mean ;) I think I've seen the
> similar with the things linked with binutils-2.29.1 too.

Haven't seen this so far, honestly.
> 
> Fencing regression tests (which reply on logging) all passed.
> 
> Pacemaker seems to work well, but every time on shutting down cluster services, crmd encounters:
> ```
> Jan 18 15:17:46 f171 crmd[30523]:     crit: GLib: Source ID 22 was not found when attempting to remove it
> ```
> 
> With the same sources but linked with binutils-2.29.1, it looks
> similar and has an additional error from crm_glib_handler:
> 
> ```
> Jan 18 15:41:13 f58.suse.de crmd[10048]:    error: crm_glib_handler: Forked child 10080 to record non-fatal assert at logging.c:73 : Source ID 23 was not found when attempting to remove it
> Jan 18 15:41:13 f58.suse.de crmd[10048]:     crit: GLib: Source ID 23 was not found when attempting to remove it
> ```

This means mixing static section callsite handling with on-the-fly
on-heap tracking, which I haven't tested but I guess I should extend
the test for this.

> The relevant code in pacemaker/lib/common/logging.c: crm_glib_handler()
> ```
>   63         glib_cs = qb_log_callsite_get(__FUNCTION__, __FILE__, "glib-handler", LOG_DEBUG, __LINE__, c
>      rm_trace_nonlog);g
>   64     }g
>   65g
>   66g
>   67     switch (msg_level) {g
>   68         case G_LOG_LEVEL_CRITICAL:g
>   69             log_level = LOG_CRIT;g
>   70g
>   71             if (crm_is_callsite_active(glib_cs, LOG_DEBUG, 0) == FALSE) {g
>   72                 /* log and record how we got here */g
>   73                 crm_abort(__FILE__, __FUNCTION__, __LINE__, message, TRUE, TRUE);g
>   74             }g
>   75             break;g
>   76g
> ```
> Not sure if it's necessarily related this change.

Yeah, strange.  Let me extend lock_test_mock.sh + surrounded files
first if it shows any new regressions wrt. on-the-fly callsites.

-- 
Jan (Poki)

jnpkrn:
In the meantime, current patchset up to 3143753 is just slightly refined version
of the previous, most of the value added comes from being strict on success
of `dl*` functions as well, which could possibly uncover some deep corner cases.

gao-yan:
FYI, the issue itself is due to double-removal of glib source:
https://github.com/ClusterLabs/pacemaker/pull/1410

krig:
Why does libqb go through all of these acrobatics at all?

jnpkrn:
On 12/02/18 11:22 +0000, Kristoffer Grönlund wrote:
> Why does libqb go through all of these acrobatics at all?

Original linker story:
https://github.com/ClusterLabs/libqb/commit/20246f544f74866885f4d9491f74152bd95e7786

Why libqb involves self-checking:
https://github.com/ClusterLabs/libqb/commit/cdb026998fa1e62cb606be048ce0a5143e6f3483

I wish things hadn't gone south in binutils, to start with.
But it happened and the fallout needs to be dealt with.

-- 
Jan (Poki)

krig:
I mean that the thing I'm trying to understand is why libqb tries to use callsite sections in order to track output filtering in the first place. To be honest, binutils doesn't look like the party at fault here, rather libqb is doing something it probably shouldn't have been doing in the first place. Was it just to avoid requiring matching push/pop calls for filter manipulation?

krig:
If that's the case, then using the GCC cleanup attribute to pop the filter option is probably a better solution.

jnpkrn:
I am not the original author but I can see that in part the intention
was to make logging as cheap and run-time seamless as possible,
which is certainly a worthy goal, given that, e.g., corosync usually
runs with real-time priority and mlock'd memory.

With this arrangement, you effectively offload page-fault prone part
of otherwise ad-hoc processing to linker at build time, and the
other, time-sensitive lookup part to early phase in run-time (upon
`qb_log_init`) -- note that everything gets resolved at once, no
matter how many log messages are there to be processed.
In my measurements with artificial, _non-favouring_ test case,
this approach is indeed still tiny bit faster.

I can only suggest reading more on the topic, the linked changes and
respective discussions, to discover that binutils indeed digressed
in handling the scaffolding this approach relies upon :-)

And no, I cannot see any way how `cleanup` attribute would help here.

Note also that callsites being tracked systemically at query-able form
within the binary means new advanced features can be introduced, such
as new super-terse "blackbox" format: #291

krig:
> And no, I cannot see any way how `cleanup` attribute would help here.

Yeah, now that I've looked at it more I see that the purpose is not entry/exit management (which `cleanup` would do), but to speed up log filtering.

Issue #286, opened 2018-01-01
[RFE] Add support for structured logging to reflect the shift towards machine log processing


jnpkrn:
This is an intended prerequisite to add support for new log back-ends like:
* native journald handover
  - https://freedesktop.org/software/systemd/man/sd_journal_send.html
  - https://freedesktop.org/software/systemd/man/systemd.journal-fields.html
* The OpenTracing Semantic Specification compliant back-end
  - https://github.com/opentracing/specification/blob/master/specification.md#the-opentracing-api
  - https://github.com/opentracing/specification/blob/master/semantic_conventions.md#standard-span-tags-and-log-fields

jnpkrn:
Possibly interesting in this context:
https://is.muni.cz/th/172673/fi_d/thesis-twoside-final-bw.pdf

jnpkrn:
For how Linux deals with this task:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/kernel/printk/printk.c?h=v4.19#n309

jnpkrn:
Ah, even RFC exists (proposed standard, exists since 2009):
https://tools.ietf.org/html/rfc5424

jnpkrn:
Regarding prior work, it's good to also mention `logfmt`, detailed,
e.g., in https://brandur.org/logfmt

jnpkrn:
See a particular motivational example:
[structured logging flag to mark secrecy-sensitive log items](https://github.com/ClusterLabs/libqb/pull/433#issuecomment-796318679)

chrissie-c:
Closing this as there has been no call for it.

PR #292, opened 2018-01-26
Allow customisable log line length

chrissie-c:
Allow the logging line length to be changed. Any reasonable length is allowable, the default is 512 as before. Anything more than 512 incurs several mallocs.

Log lines that exceed the length will have ... as the last 3 characters.

### DIFF
--- a/include/qb/qblog.h
+++ b/include/qb/qblog.h
@@ -584,7 +584,7 @@ enum qb_log_conf {
 	QB_LOG_CONF_EXTENDED,
 	QB_LOG_CONF_IDENT,
 	QB_LOG_CONF_MAX_LINE_LEN,
-	QB_LOG_CONF_ELIPSIS,
+	QB_LOG_CONF_ELLIPSIS,
 };
 
 enum qb_log_filter_type {
--- a/include/qb/qblog.h
+++ b/include/qb/qblog.h
@@ -584,6 +584,7 @@ enum qb_log_conf {
 	QB_LOG_CONF_EXTENDED,
 	QB_LOG_CONF_IDENT,
 	QB_LOG_CONF_MAX_LINE_LEN,
+	QB_LOG_CONF_ELIPSIS,
 };
 
 enum qb_log_filter_type {
--- a/include/qb/qblog.h
+++ b/include/qb/qblog.h
@@ -265,6 +265,7 @@ extern "C" {
 #define LOG_TRACE    (LOG_DEBUG + 1)
 
 #define QB_LOG_MAX_LEN 512
+#define QB_LOG_ABSOLUTE_MAX_LEN 4096
 #define QB_LOG_STRERROR_MAX_LEN 128
 
 typedef const char *(*qb_log_tags_stringify_fn)(uint32_t tags);
--- a/include/qb/qblog.h
+++ b/include/qb/qblog.h
@@ -582,6 +582,7 @@ enum qb_log_conf {
 	QB_LOG_CONF_FILE_SYNC,
 	QB_LOG_CONF_EXTENDED,
 	QB_LOG_CONF_IDENT,
+	QB_LOG_CONF_MAX_LINE_LEN,
 };
 
 enum qb_log_filter_type {
--- a/lib/log.c
+++ b/lib/log.c
@@ -1222,7 +1222,12 @@ qb_log_ctl2(int32_t t, enum qb_log_conf c, qb_log_ctl2_arg_t arg_not4directuse)
 		conf[t].extended = arg_i32;
 		break;
 	case QB_LOG_CONF_MAX_LINE_LEN:
-		conf[t].max_line_length = arg_i32;
+		/* arbitrary limit, but you'd be insane to go further */
+		if (arg_i32 > QB_LOG_ABSOLUTE_MAX_LEN) {
+			rc = -EINVAL;
+		} else {
+			conf[t].max_line_length = arg_i32;
+		}
 		break;
 
 	default:
--- a/lib/log.c
+++ b/lib/log.c
@@ -193,16 +193,21 @@ _cs_matches_filter_(struct qb_log_callsite *cs,
 #endif
 #endif
 static inline void
-cs_format(char *str, struct qb_log_callsite *cs, va_list ap)
+cs_format(char *str, size_t maxlen, struct qb_log_callsite *cs, va_list ap)
 {
 	va_list ap_copy;
 	int len;
 
 	va_copy(ap_copy, ap);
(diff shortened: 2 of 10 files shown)


lib/log_blackbox.c
```diff
@@ -110,11 +110,11 @@ _blackbox_vlogger(int32_t target,
...
-	msg_len = qb_vsnprintf_serialize(chunk, QB_LOG_MAX_LEN, cs->format, ap);
-	if (msg_len >= QB_LOG_MAX_LEN) {
+	msg_len = qb_vsnprintf_serialize(chunk, max_size, cs->format, ap);
+	if (msg_len >= max_size) {
 	    chunk = msg_len_pt + sizeof(uint32_t); /* Reset */
 
-	    msg_len = qb_vsnprintf_serialize(chunk, QB_LOG_MAX_LEN,
+	    msg_len = qb_vsnprintf_serialize(chunk, max_size,
 		"Log message too long to be stored in the blackbox.  "\
 		"Maximum is QB_LOG_MAX_LEN" , ap);
```
jnpkrn: Not changing the message indicates me there are likely other issues to be hit,
at least in blackbox context.

lib/log.c
```diff
@@ -1196,6 +1221,9 @@ qb_log_ctl2(int32_t t, enum qb_log_conf c, qb_log_ctl2_arg_t arg_not4directuse)
 	case QB_LOG_CONF_EXTENDED:
 		conf[t].extended = arg_i32;
 		break;
+	case QB_LOG_CONF_MAX_LINE_LEN:
+		conf[t].max_line_length = arg_i32;
+		break;
```
jnpkrn: No boundary requirement is no-go.

jnpkrn: You could, e.g., easily overflow the blackbox with a single message like that.

chrissie-c:
TBH I'm tempted to leave the blackbox line length restriction at QB_LOG_MAX_LEN as it's the one place where having log lines can really clog up the system. 

Any ideas for a limit for what would be the actual hard limit imposed in qb_log_ctl()? In principle I was hoping to avoid arbitrary limits in there. if you ask for more than is reasonable then you suffer the consequences at malloc() time - same as if you called malloc() yourself. I'm easy on this though, a 64K limit would not be unreasonable I think.

kgaillot:
> Any ideas for a limit for what would be the actual hard limit imposed in qb_log_ctl()? In principle I was hoping to avoid arbitrary limits in there. if you ask for more than is reasonable then you suffer the consequences at malloc() time - same as if you called malloc() yourself. I'm easy on this though, a 64K limit would not be unreasonable I think.

No preference here. I think an upper limit anywhere from 2K to 16K would be reasonable.

jnpkrn:
On 30/01/18 17:53 +0000, Ken Gaillot wrote:
>> Any ideas for a limit for what would be the actual hard limit
>> imposed in qb_log_ctl()? In principle I was hoping to avoid
>> arbitrary limits in there. if you ask for more than is reasonable
>> then you suffer the consequences at malloc() time - same as if you
>> called malloc() yourself. I'm easy on this though, a 64K limit
>> would not be unreasonable I think.
> 
> No preference here. I think an upper limit anywhere from 2K to 16K
> would be reasonable.

I'd stay with 4K (page size) as the all-time maximum for _the whole,
combined, final message_ (incl. trailing '\0', giving recommendation
not to exceed 2K, and even better to fit into default 512B as it's
a pain to deal with longish lines -- and for machine handling, it's
becoming prevalent to use structured logging, see #286) limit + would
be noisy if 512B limit for a _format string_ for internally logged
messages is exceeded -- QB_LOG_FILTER_FORMAT_REGEX filter would
presumably suffer a lot performance-wise if we were liberal here as
well (we already are, but since the limited message length so far, it
would be pretty stupid idea ... until the point of relaxing the
maximum overall length), and frankly, if 512 ASCII characters are not
enough to explain some matter incl. references, something is utterly
wrong (plus emergency workaround is as simple as using qb_log("%s",
actual_string)).  Since the format string is necessarily a
compile-time constant (you cannot build otherwise with callsite
section enabled), it should be feasible to arrange such a check as
assurably compile-time assertion.

And going deeper, proposed simple API is insufficient.

What one likely wants to know: "under current conditions, with the
current target->format (possibly modified qb_log_format_set()),
_user_tags_stringify_fn etc. how many bytes are there left for my
actual messages?".

Hence, qb_log_ctl should return this information upon setting
new value (0 ~ do not change anything, just return this
information).  That's also rather convoluted, e.g. for %g extra
specifier without width constraint, "tags" for all 32+1 exclusive
bit combinations should be obtained as a best effort heuristics
which one yields longest string (and documention made explicit
it's desirable to have the specifier always width-constrained).

It'd be then easy to adapt new size accordingly if one is not
satisfied, by mere subtraction, and ideally it would happen
only in corner case scenarios (excessively long hostname?)

Also, figuring this limit internally, it'd be easier (worst-case
estimation and possibly slight correction) to implement strace-style
information on how many bytes have been omitted, which is definitely
more informational than mere three dots (which cannot really be
discerned from regular ellipsis in some conceivable cases).  Related
extension would be to add ellipsis only for the actual inner message,
not at the end of the buffer, should the fields be ordered in fancy
ways.

-- 
Jan (Poki)

include/qb/qblog.h
```diff
@@ -584,6 +584,7 @@ enum qb_log_conf {
 	QB_LOG_CONF_EXTENDED,
 	QB_LOG_CONF_IDENT,
 	QB_LOG_CONF_MAX_LINE_LEN,
+	QB_LOG_CONF_ELIPSIS,
```
kgaillot: fyi spelling is "ellipsis"

chrissie-c: Thanks, I knew it looked wrong, but couldn't quite figure out why!

jnpkrn:
In retrospect, thanks for sticking with my advice to keep the maximum
limited in a reasonable way (4KiB here), see also glibc's extraneously
imposed allocations:
https://gcc.gnu.org/ml/gcc-patches/2016-08/msg01758.html
that I believe would kick in for `vsnprintf` in `qb_log` call
stack as well.

(from the neighbour context
https://gcc.gnu.org/ml/gcc-patches/2016-08/msg01772.html
from the neighbour context
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=88993
related to upcoming GCC 9)

chrissie-c:
Good spot! it just seemed a sensible suggestion at the time :)

jnpkrn:
Oh, and this likely requires even more considerations before being
ready for prime time (bigger messages means more likeliness of
non-atomic appends to the file):

https://lists.clusterlabs.org/pipermail/users/2019-April/025735.html

PR #295, opened 2018-02-21
Hurd support

wferi:


chrissie-c:
Do you still want this PR? I think it'll need updating if you do :)

PR #296, opened 2018-02-21
Fix comment typo

wferi:
Or there may be an explanation instead: according to socket(7), the
Linux kernel doubles the set value "to allow space for bookkeeping
overhead [...] and this doubled value is returned by getsockopt(2)."
In other words, this really is some kernel buffer size, not the size
of the biggest packet possible to send (as far as I understand it).
Note that BSD does not do this doubling.

Signed-off-by: Ferenc Wágner <wferi@debian.org>


wferi:
I will do that.
Shouldn't the code change as well to account for the bookkeeping overhead on BSD? Or does BSD not need extra buffer space?

chrissie-c:
Thanks.

It's a fair point about BSD and the true answer is that I don't know, but we haven't had any problems reported. So maybe it does its own bookkeeping behind the scenes?

wferi:
@chrissie-c , why did you close this PR?

chrissie-c:
Total mistake, please accept my apologies!

wferi:
I'll put the socket buffer stuff into a new PR and keep this one for typos.

chrissie-c:
Fair enough! Are there others to come is is this ready for pulling?

wferi:
Sorry for the delay! I haven't got more typo fixes queued at the moment. Shall I rebase this to current master?

chrissie-c:
Yes please, then I'll pull it and we can get on with our lives :)

wferi:
Done.

chrissie-c:
Thanks!

PR #298, opened 2018-03-02
tests: Improve test isolation

chrissie-c:
Make all the IPC tests run with a common date/pid stamp name, so that
the final resource.test only fails if it finds one of OUR files left
lying around and not those from another test.

### DIFF
--- a/tests/check_ipc.c
+++ b/tests/check_ipc.c
@@ -122,26 +122,19 @@ exit_handler(int32_t rsignal, void *data)
 static void
 set_ipc_name(const char *prefix)
 {
-	/* We have to make the server name as unique as possible given
-	 * the build- (seconds part of preprocessor's timestamp) and
-	 * run-time (pid + lower 16 bits of the current timestamp)
-	 * circumstances, because some build systems attempt to generate
-	 * packages for libqb in parallel.  These unit tests are run
-	 * during the package build process.  2+ builds executing on
-	 * the same machine (whether containerized or not because of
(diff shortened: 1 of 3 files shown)


tests/check_ipc.c
```diff
@@ -122,26 +122,19 @@ exit_handler(int32_t rsignal, void *data)
...
-	if (t_sec[0] == '\0') {
-		const char *const found = strrchr(__TIME__, ':');
-		strncpy(t_sec, found ? found + 1 : "-", sizeof(t_sec) - 1);
-		t_sec[sizeof(t_sec) - 1] = '\0';
+	FILE *f;
+	char process_name[256];
+
+	/* The process-unique part of the IPC name has already been decided
+	 * and stored in the file ipc-test-name.
+	 */
```
jfriesse: It may be nice to have some fallback when `ipc-test-name` file does not exists because right now, result is going to be random bytes from memory.

tests/start.test
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+#
+# Generate a unique(ish) name for the IPCs we will use in the tests and
+# save it in a file for all of the tests to use. This way we know for sure
+# which sockets are our and which we can ignore.
+# The test programs all add "qb-test-<name>-" to the front of this.
+#
+NAME="$$-`date +%N`"
+echo -n "$NAME" > ipc-test-name
```
jfriesse: New line missing

jfriesse:
Perfect, looks good now (and ci passed :) ), so ACK

Issue #299, opened 2018-03-04
linker command fails on openbsd


jirib:
```
$ sysctl kern.version                                                                                                                                                                                        
kern.version=OpenBSD 6.3-beta (GENERIC) #25: Fri Mar  2 22:51:43 MST 2018
    deraadt@amd64.openbsd.org:/usr/src/sys/arch/amd64/compile/GENERIC
```

Trying to build 1.0.3:

```
...
libtool: compile:  cc -DHAVE_CONFIG_H -I. -I../include -I../include/qb -I../include -I../include -O2 -pipe -Wall -Wextra -Wunused -Wshadow -Wmissing-prototypes -Wmissing-declarations -Wstrict-prototypes -Wpointer-arith -Wwrite-strings -Wcast-align -Wbad-function-cast -Wmissing-format-attribute -Wfloat-equal -Wformat=2 -Woverlength-strings -Winit-self -Wuninitialized -Wunknown-pragmas -Wno-unused-parameter -Wno-format-nonliteral -Wno-sign-compare -MT strchrnul.lo -MD -MP -MF .deps/strchrnul.Tpo -c strchrnul.c  -fPIC -DPIC -o .libs/strchrnul.o
libtool: compile:  cc -DHAVE_CONFIG_H -I. -I../include -I../include/qb -I../include -I../include -O2 -pipe -Wall -Wextra -Wunused -Wshadow -Wmissing-prototypes -Wmissing-declarations -Wstrict-prototypes -Wpointer-arith -Wwrite-strings -Wcast-align -Wbad-function-cast -Wmissing-format-attribute -Wfloat-equal -Wformat=2 -Woverlength-strings -Winit-self -Wuninitialized -Wunknown-pragmas -Wno-unused-parameter -Wno-format-nonliteral -Wno-sign-compare -MT strchrnul.lo -MD -MP -MF .deps/strchrnul.Tpo -c strchrnul.c -o strchrnul.o >/dev/null 2>&1
cc -E -xc -I../include -D_GNU_SOURCE -C -P qblog_script.ld.in \
  | sed -n "/$(sed -n '/^[^#]/{s/[*\/]/\\&/g;p;q;}' qblog_script.ld.in)/,$ p" \
  > qblog_script.ld
[...]

PR #303, opened 2018-03-20
Fix distcheck

chrissie-c:
Miscellaneous fixes and cleanups to get 'make distcheck' to work on most platforms.

There's a thing in here to speed up the FreeBSD IPC stress tests. I know it means that the BSD test is no longer really a stress test, but it can take over an hour to run with 70000 iterations and that;s just getting in everyone's way.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -37,6 +37,9 @@ MAINTAINERCLEANFILES	= Makefile.in aclocal.m4 configure depcomp \
 
 ACLOCAL_AMFLAGS		= -I m4
 
+# For test suite
+export SOCKETDIR
+
 dist_doc_DATA		= COPYING INSTALL README.markdown
 
 SUBDIRS			= include lib docs tools tests examples
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -137,6 +137,8 @@ install-exec-hook: qblog_script.ld
 	  sed -i -- "s/libqb.so.<digit>/$${so_ver}/" \
 	    "$(DESTDIR)$(libdir)/libqb.so-t" "$(DESTDIR)$(pkgconfigexecdir)/libqb.pc"
 	mv -f "$(DESTDIR)$(libdir)/libqb.so-t" "$(DESTDIR)$(libdir)/libqb.so"
+	rm -f "$(DESTDIR)$(pkgconfigexecdir)/libqb.pc--"
+	rm -f "$(DESTDIR)$(libdir)/libqb.so-t--"
 endif
 
 if HAVE_SPLINT
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -80,8 +80,9 @@ check: check-headers
 .PHONY: check-headers
 check-headers: $(auto_c_files:.c=.o)
 
-distclean-compile:
+distclean-local:
 	rm -rf auto_*.c
+	rm -rf .deps
 
 if HAVE_DICT_WORDS
 if HAVE_SLOW_TESTS
--- a/tests/check_ipc.c
+++ b/tests/check_ipc.c
@@ -23,6 +23,7 @@
 
 #include "os_base.h"
 #include <sys/wait.h>
+#include <sys/un.h>
 #include <signal.h>
 
 #include "check_common.h"
@@ -132,7 +133,6 @@ set_ipc_name(const char *prefix)
 	if (f) {
 		fgets(process_name, sizeof(process_name), f);
 		fclose(f);
-		fprintf(stderr, "CC: using stored SHM name %s\n", process_name);
 		snprintf(ipc_name, sizeof(ipc_name), "%s%s", prefix, process_name);
 	} else {
 		/* This is the old code, use only as a fallback */
@@ -145,7 +145,6 @@ set_ipc_name(const char *prefix)
 
 		snprintf(ipc_name, sizeof(ipc_name), "%s%s%lX%.4x", prefix, t_sec,
 			 (unsigned long)getpid(), (unsigned) ((long) time(NULL) % (0x10000)));
-		fprintf(stderr, "CC: using calculated SHM name %s\n", ipc_name);
 	}
 }
 
@@ -1436,11 +1435,15 @@ END_TEST
 START_TEST(test_ipcc_truncate_when_unlink_fails_shm)
 {
 	char sock_file[PATH_MAX];
+	struct sockaddr_un socka;
+
 	qb_enter();
 	ipc_type = QB_IPC_SHM;
 	set_ipc_name(__func__);
 
 	sprintf(sock_file, "%s/%s", SOCKETDIR, ipc_name);
+	sock_file[sizeof(socka.sun_path)] = '\0';
+
 	/* If there's an old socket left from a previous run this test will fail
 	   unexpectedly, so try to remove it first */
 	unlink(sock_file);
--- a/tests/functional/GNUmakefile
+++ b/tests/functional/GNUmakefile
@@ -10,7 +10,7 @@ distclean maintainer-clean:
 	# includes, which are swiped when processing one subdir and then
 	# missing, as a hard error, for the other)
 	@$(MAKE) -C log_external $@
-	@mkdir .deps
+	@mkdir -p .deps
 	@touch .deps/log_client.Po .deps/log_interlib.Plo .deps/log_interlib_client.Po
 	@$(MAKE) -C log_internal $@
 	@$(MAKE) -f Makefile $@ SUBDIRS=
--- a/tests/resources.test
+++ b/tests/resources.test
@@ -2,7 +2,7 @@
 RETURN=0
 
 IPC_NAME=`cat ipc-test-name 2>/dev/null`
-for d in /dev/shm /var/run; do
+for d in /dev/shm /var/run $SOCKETDIR; do
 	leftovers=$(find $d -name qb-test*${IPC_NAME}* -size +0c 2>/dev/null | wc -l)
 	if [ "${leftovers}" -gt 0 ]; then
 		echo
@@ -30,4 +30,7 @@ if [ $? -eq 0 ]; then
 	RETURN=1
 fi
 
+# Keep it tidy - distcheck checks we have not left a mess
+rm -f ipc-test-name
+rm -f crash_test_dummy.core
 exit $RETURN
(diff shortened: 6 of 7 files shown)


jnpkrn:
There's a lot of context and some alignment with the existing checks
missing.

Under which circumstances does the check take 60+ minutes and does this
anomaly happens only with FreeBSD?

Just tried, "make check" incl. compilation took about 8 minutes,
FreeBSD 11.1, virtualized under KVM hypervisor, single CPU allocated
from 8 of i7-6820HQ.

Makefile.am
```diff
@@ -37,6 +37,10 @@ MAINTAINERCLEANFILES	= Makefile.in aclocal.m4 configure depcomp \
 
 ACLOCAL_AMFLAGS		= -I m4
 
+DISTCHECK_CONFIGURE_FLAGS = --with-socket-dir=/tmp/libqb-$(VERSION)-$(shell echo $$$$)
```
jnpkrn: Only to be removed with subsequent commit?

tests/functional/GNUmakefile
```diff
@@ -10,7 +10,7 @@ distclean maintainer-clean:
 	# includes, which are swiped when processing one subdir and then
 	# missing, as a hard error, for the other)
 	@$(MAKE) -C log_external $@
-	@mkdir .deps
+	@mkdir .deps 2>/dev/null || true
```
jnpkrn: You likely want `mkdir -p` instead of suppressing permission issues etc.

chrissie-c: good point

Makefile.am
chrissie-c: Yes, I decided it was better done in the CI itself rather than hard-coding it in the Makefile.

chrissie-c:
I am no FreeBSD expert. All I know is that all the FreeBSD VMs I have ever tried this on (the 3 I have here on two different machines) and the one in the Jenkins CI all take around an hour to run. If  a FreeBSD expert wants to fix this then it's fine by me, but I don't have the time to do it.

jnpkrn:
Wrt. alignment with the existing checks, there's `--enable-slow-tests`
to `configure` script.  We can either reuse that, or perhaps better yet,
apply a similar logic with optional `--disable-sticky-tests` (work name).
As demonstrated, it should not be hard-wired to apply on FreeBSD
unconditionally because there are cases with this system where the
testsuite runs reasonably out-of-the-box.

chrissie-c:
Either of those is fine by me - and certainly tidier than what I had in there. My main aim here is just to get the stupid thing working :)

lib/Makefile.am
```diff
@@ -137,6 +137,8 @@ install-exec-hook: qblog_script.ld
 	  sed -i -- "s/libqb.so.<digit>/$${so_ver}/" \
 	    "$(DESTDIR)$(libdir)/libqb.so-t" "$(DESTDIR)$(pkgconfigexecdir)/libqb.pc"
 	mv -f "$(DESTDIR)$(libdir)/libqb.so-t" "$(DESTDIR)$(libdir)/libqb.so"
+	rm -f "$(DESTDIR)$(pkgconfigexecdir)/libqb.pc--"
+	rm -f "$(DESTDIR)$(libdir)/libqb.so-t--"
 endif
```
jnpkrn: Took me some moments to realize the issue -- there's an ambiguity
across systems as to whether `--` should be considered as an argument
to `-i` option or whether this option should be considered orphan (not
accompanied with any argument as it's optional, anyway), followed with
`delimiter indicating the end of options`:
http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html#tag_12_02

tests/Makefile.am
```diff
@@ -82,6 +82,7 @@ check-headers: $(auto_c_files:.c=.o)
 
 distclean-compile:
 	rm -rf auto_*.c
+	rm -rf .deps
```
jnpkrn: Looks to me the target should rather be named `distclean-local`,
cannot find any support for this one to be supported, hence we
should avoid surprises with the gray zone.

Makefile.am
jnpkrn: But ... introducing temporary, unjustified inter-patchset noise is not
nice should anyone care about history, bisecting etc.

lib/Makefile.am
chrissie-c: Do we have a better answer to this other than just removing the junk files? That chunk of shell code is extremely opaque (I remember commenting on that at the time.)

Makefile.am
chrissie-c: It would have disappeared when I committed the final patch.

tests/Makefile.am
chrissie-c: interesting! All these built-in destinations still throw me, and some are hard to find in the doc. disclean-local works for me so I''m happy to change it.

Is there a good reference for these things? the info files seem to be unhelpful.

jnpkrn: https://www.gnu.org/software/automake/manual/automake.html#Clean
(as applicable to many other cases incl. pacemaker: single-page compilation,
finding through the page, even though info pages map 1:1 for most GNU
projects, IIRC)

Makefile.am
jnpkrn: I thought the typical approach is to evolve the patchset towards the final
form directly as a part of the review.  I see you want to see the response
from the CI, but cannot you prototype using your fork for that?

tests/check_ipc.c
```diff
@@ -64,7 +64,7 @@ static const int MAX_MSG_SIZE = DEFAULT_MAX_MSG_SIZE;
 #define GIANT_MSG_DATA_SIZE MAX_MSG_SIZE - sizeof(struct qb_ipc_response_header) - 8
 
 /* This can take AGES (like, an hour) on FreeBSD */
-#ifdef __FreeBSD__
+#ifndef HAVE_SLOW_TESTS
```
jnpkrn: I think we should relax testing being in place for several years already
only on demand, through that sketched `--disable-sticky-tests` equivalent.
The hypothesis that it's _always_ slow on FreeBSD has been invalidated,
so only a very very specific subset of all setups is affected (VMs with
FreeBSD lacking virtio support under KVM?).

Makefile.am
chrissie-c: TBH I just want something that works. I hate working with autotools and it's ghastly friends. Every keypress I make regarding them makes me less and less motivated.

chrissie-c:
We don' t know whether it's a small or a large subset of FreeBSD installations that are slow - we have 3 data points. As it stands (because it affects the ones I'm using), I'm more inclined to put the original #ifdef __FreedBSD__ back in (I agree, disabling it on Linux is a a bad move) unless someone has a proper fix or can explain why all all my VMs (and  Fabio's) exhibit the problem and how to fix it.

jnpkrn:
Can you at least collect which FreeBSD versions are you talking about?
`freebsd-version` should do.  Point is, if they are rather outdated, I guess
we should not care at all in the build not instructed otherwise.

fabbione:
On 3/22/2018 3:44 PM, Chrissie Caulfield wrote:
>
> We don' t know whether it's a small or a large subset of FreeBSD
> installations that are slow - we have 3 data points. As it stands
> (because it affects the ones I'm using), I'm more inclined to put the
> original #ifdef *FreedBSD* back in (I agree, disabling it on Linux is
> a a bad move) unless someone has a proper fix or can explain why all
> all my VMs (and Fabio's) exhibit the problem and how to fix it.
>

I am a bit lost on what the original issue was, but to be honest, a test
suite should work regardless if the VM is slow or fast. The test suite
passes on sparc64 that is 100x slower than any of those VMs, so the
problem might be real somewhere else.

Fabio

chrissie-c:
Both the slow FreeBSD VMs are 11.1 and were recently installed. They are also on three different hosts - though all using KVM as the virtualisation platform. I really don't want to get stuck on this. I'm happy to take the ifdef out completely if it annoys people.

chrissie-c:
OK. I just moved my FreeBSD VM to another host and the full 70000 iteration tests completed in 8 minutes, same as yours Poki, so I'm happy it's a hosting problem. from that I suspect that people running it on bare metal are probably not seeing the problem and I'm going to remove that ifdef completely.

jnpkrn:
Sounds like it may actually be the host the causes this (either with dated
kernel/KVM and or being virtualized on its own, i.e., nested virtualization?).

chrissie-c:
7.4 kernels are slow here too. (no nested virt - that's insane). Fedora 27 kernel seems to run reasonably though

jfriesse:
@chrissie-c Patchset looks ok for me, so ACK from my side.

chrissie-c:
Thanks. I appreciate that the bit around the .ld file generation is suboptimal but we can fix that later if needs be. My main aim here is to get the tests back working.

lib/Makefile.am
jnpkrn: Let me include the fix for that into #288 I am going to overhaul a bit
(incl. a solution for backdated OpenBSD per #299).  We need to get
rid of `sed -i` as it's not POSIX :-/

kgaillot: Ah, don't get me started ... spent way too much time yesterday trying to come up with a portable function to replace sed -i and gave up -- it's not impossible, but far uglier than this simple workaround:

    sed ... "$INFILE" > "${INFILE}.$$"
    mv -- "${INFILE}.$$" "$INFILE"

It obviously only works on one input file at a time. If you need to preserve permissions, you have to do this first:

    cp -p "$INFILE" > "${INFILE}.$$"

(A portable way to preserve permissions without copying the entire file is also too ugly for words.)

jnpkrn: If you had suspicion what triggered this response of mine, you were right,
I came by that pacemaker commit, thanks for insights.

chrissie-c:
Update on the 'slow BSD' problem - it's a Linux kernel bug and is fixed in the RHEL7.5 kernel where the BSD tests take 8 minutes again.

jnpkrn:
Thanks for the upgrade, resisting the immediate urge paid off this time.

fabbione:
We will upgrade the CI infrastructure probably next week as 7.5 just went GA and we need to plan downtime. The distcheck is temporary disabled for BSD for this few days we need for preparation.

PR #306, opened 2018-04-05
[WIP] harden printf-format checking at the clients

jnpkrn:
Update: now in substantially better shape, though I want to add new
validation it indeed works somewhere under `tests/functional`
(again, this is rather specialized test to be triggered manually
so as to see whether a compiler indeed complains).

### DIFF
--- a/include/qb/qblog.h
+++ b/include/qb/qblog.h
@@ -36,8 +36,9 @@ extern "C" {
 #include <errno.h>
 #include <syslog.h>
 #include <string.h>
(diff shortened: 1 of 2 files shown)


include/qb/qblog.h
```diff
@@ -500,7 +504,7 @@ void qb_log_from_external_source_va(const char *function,
  * @param fmt usual printf style format specifiers
  * @param args usual printf style args
```
jnpkrn: Ah, documentation for the now not manifested arguments.

PR #307, opened 2018-04-10
ipc_shm: Don't truncate SHM files of an active server

chrissie-c:
I've put in an extra check so that clients don't truncate the
SHM file if the server still exists. Sadly on FreeBSD we can't
get the server PID for the client (unless someone has a patch handy!)
so we still do the truncate when disconnected. As a backstop (and also
to cover the BSD issue) I've added a SIGBUS trap to the server shutdown
so that it doesn't cause a server crash.

### DIFF
--- a/include/qb/qbipcs.h
+++ b/include/qb/qbipcs.h
@@ -142,6 +142,10 @@ typedef void (*qb_ipcs_connection_created_fn) (qb_ipcs_connection_t *c);
  * successfully created.
  * @note if you return anything but 0 this function will be
  * repeatedly called (until 0 is returned).
+ *
+ * With SHM connections libqb will briefly trap SIGBUS during the
+ * disconnect process to guard against server crashes if the mapped
+ * file is truncated. The signal will be restored afterwards.
  */
 typedef int32_t (*qb_ipcs_connection_closed_fn) (qb_ipcs_connection_t *c);
 
--- a/lib/ipc_int.h
+++ b/lib/ipc_int.h
@@ -92,6 +92,7 @@ struct qb_ipcc_connection {
 	char name[NAME_MAX];
 	int32_t needs_sock_for_poll;
 	gid_t egid;
+	pid_t server_pid;
 	struct qb_ipc_one_way setup;
 	struct qb_ipc_one_way request;
 	struct qb_ipc_one_way response;
--- a/lib/ipc_setup.c
+++ b/lib/ipc_setup.c
@@ -494,6 +494,7 @@ qb_ipcc_us_setup_connect(struct qb_ipcc_connection *c,
 
 	qb_ipc_auth_creds(data);
 	c->egid = data->ugp.gid;
+	c->server_pid = data->ugp.pid;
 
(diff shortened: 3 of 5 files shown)


chrissie-c:
Good point about the signal. I'll use sigaction and restore the original signal handler on exit.

jnpkrn:
I am thinking what other/out there solutions are possible.
I agree mangling with signals behind the client's back feels
dirty and backward.  Client should only proceed if the server
is dead.  Would some kind of locking or the semaphore
mechanism (presumably auto-cleaned when server crashes)
that's already in use help?

Tangentially, regarding robustness (but that may be overboard
paranoid), what if the client accidentally calls `ftruncate(fd, 0)`
where `fd` was evaluated in a flawed way (there's no encapsulation
or namespacing within program space when it comes to these
descriptors), hence not shooting it's own managed descriptor,
but rather that of libqb?

chrissie-c:
I expected that bit of code to be controversial, but I expected my use of longjmp to have generated more heat! As the main problem seems to be that BSD can't tell if the server has gone away I might look closer into seeing how viable it is to get the server PID on FreeBSD.  A shame we can't pass it in the shared memory segment but that would break compatibility

jfriesse:
Real Programmers aren't afraid to use GOTO's. ;)

chrissie-c:
I've been thinking about other ways to protect the SHM segment. I like the idea of a semaphore, but that breaks backwards compatibilty between client & server so would have to be a v2.0 feature sadly. Guarding against a truncate that happens earlier than at client shutdown would also be nice but would also require some serious revamping of the IPC mechanism I suspect. I don't regard that as a huge problem at the moment as clients are authenticated at connect time.

There is a potential for a race with the PID checking I have in this PR, though it is tiny because the time between a client and server quitting is going to be very small number of schedules. There is a way to get the pid of a server on FreeBSD but I need to check it won't break backwards compatibility as it requires the server to send a message to the client.

chrissie-c:
Looking into signal ignoring - libqb IPC spends a lot of time flipping SIGPIPE on and off.

jnpkrn:
Cautiousness appreciated, though I checked this should not be of a concern
with neither Linux nor FreeBSD (and respective libc environments),
former responds to `MSG_NOSIGNAL` flag to `send` family of functions,
the latter also honours `SO_NOSIGPIPE` socket option.  Also checked that
`qb_sigpipe_ctl` with the former boils down to mere `retq` on amd64 arch.
We could help the optimization further when declaring that function `inline`:
#308

chrissie-c:
I haven't managed to get a backward-compatible way of getting the server PID in FreeBSD sadly. I suspect it's possible but I don't want to waste any more time on it. If we can't do that and don't want to ignore SIGBUS during IPC shutdown then I suppose the alternative is not to do the truncate on FreeBSD.

jfriesse:
@chrissie-c At least for me the solution you've sent https://github.com/ClusterLabs/libqb/pull/307/commits/75468f7ad5307a7942ef264384d2b3383cd7a42d is just fine.

PR #308, opened 2018-04-18
Low: qb_{sigpipe_ctl,socket_nosigpipe}: optimize through inlining

jnpkrn:
It was pointed out by Chrissie that the former function, qb_sigpipe_ctl,
can be a bottle neck when SIGPIPE handling is being temporarily
ignored around UNIX socket IO (which also makes libqb generally
thread-unfriendly!), unless the platform provides better means of
achieving that effect, in which case the function is no-op.
Due to multiple callers, it would be impractical to have such
a build-time conditionalization at the callsites rather than at that
function itself, but with Linux and FreeBSD (amongst possibly more)
this leads to entering completely empty frames needlessly.  Solution is
to treat the function as inline at the affected callsites, which feeds
the compiler's optimizing phases as expected.  Nonetheless, to retain
(API-bypassing) ABI compatibility, we also preserve said function
+ another falling into the same category, qb_socket_nosigpipe,
also as externally visible symbols.


jnpkrn:
Likely due to measurement error, mean wall clock time of `tests/ipc.test`
was one second more with this change (0.3% from ~270 seconds),
however mean `rss` was jumping before the change (for no apparent
reason) and the mean value was higher.  All in 3 repeated runs and
using `-O3` for the library.  More representative benchmark, both
code-wise and iteration-wise is needed.

chrissie-c:
The idea seems sound but the increase in time worries me. Also -O2 is generally the more common optimisation level rather than -O3 so testing with this would be more representative of actual 'in the field' performance. Though adding -finline-functions would probably be needed.

jnpkrn:
And per Travis, apparently Clang is not GCC-compatible (may be just a matter
of explicitly passing `-std=c99`, though).

jnpkrn:
Ok, clang is just more sensitive wrt. "once inline, always inline".

chrissie-c:
Thanks for doing the research on this. It just goes to show that not all optimisations are worth doing :)

jnpkrn:
It just demonstrated no immediate reliable benchmarkability of the
observation related to your worries about constantly context switching
to kernel wrt. signal handling, which thankfully is not happening with
the main platforms.

Currently, there's a significant gap in performance self-testing of libqb,
assuming any slight decrease may impact the cluster stack heavily
(and vice versa, unlocking new speed improvements).

PR #310, opened 2018-05-10
log: Use RTLD_NOLOADwhen checking symbols

chrissie-c:
On FreeBSD 11 call dlopen on a shared library causes the constructors
to run again. As we're just getting symbols we don't need this to
happen.

Actually we don't WANT it to happen because it can cause qb_log_init to
be called twice (recursively) and the dlnames list gets corrupted. This
causes corosync (at least) to crash at startup.

### DIFF
--- a/lib/log.c
+++ b/lib/log.c
@@ -831,7 +831,7 @@ _log_so_walk_dlnames(void)
 	qb_list_for_each_safe(iter, next, &dlnames) {
 		dlname = qb_list_entry(iter, struct dlname, list);
 
-		handle = dlopen(dlname->dln_name, RTLD_LAZY);
+		handle = dlopen(dlname->dln_name, RTLD_LAZY|RTLD_NOLOAD);
 		error = dlerror();
 		if (!handle || error) {
 			qb_log(LOG_ERR, "%s", error);


kgaillot:
I'm not very familiar with the dl*() functions. Doesn't the library need to be loaded to use dlsym()? It looks like RTLD_NOLOAD is a glibc thing, I guess we don't support anything else?

chrissie-c:
I think by 'load' it means activate ready for execution - which is the usual use for dlopen of course. It certainly means the constructors are called (which is the problem here). with NOLOAD I suspect it just loads it into memory so you can get the symbols from it. But because we are already walking the tree of loaded libraries we don't need to 'LOAD' them again, the symbols are already there (I have checked this).

It's certainly supported by FreeBSD (which is where I see the problem) and Linux, but if you think it might be a problem I can set it at ./configure time.

chrissie-c:
We're not looking for the whole symbol table, just  for the logging segments inside each loaded object so I think dlsym() is fine. Also adding NOLOAD removes the potential for the race as will only look inside the objects that are already loaded in memory and in use by the program - which is what we want. 

If the shared object is replaced we most certainly do NOT want the new symbols as they might not be correct for what is in RAM.

wferi:
Sure, but with `RTLD_NOLOAD` added `dlopen()` will return `NULL` if the shared object is replaced (even though its previous instance is already loaded), thus you won't be able to use `dlsym()`.

chrissie-c:
TBH that's probably preferable to getting the wrong symbol, which would be a guaranteed segfault :)

wferi:
Probably so, though I'm not sure what happens on symbol lookup failure.
And I guess it would be best to get the correct symbol anyway.

chrissie-c:
It's a messy area I agree, thanks for your input. I'll see if anyone else has any comments before I commit it. Just to be on the safe side :)

wferi:
To be explicit, I support merging this PR. We're probably better off with it, even though the code might not be perfect still. I'm unsure because the quite insightful https://github.com/ClusterLabs/libqb/pull/266#issuecomment-383072666 and its followups do not mention this supposed race.

Issue #237, opened 2016-11-28
blackbox corruption protection insufficient


jnpkrn:
See
https://travis-ci.org/jnpkrn/libqb/jobs/178981147#L1766

```
 ==== Corrupt blackbox test 31/32 ====
Dumping the contents of butchered_blackbox.fdata
[debug] shm size:8192; real_size:8192; rb->word_size:2048
[debug] read total of: 8212
Ringbuffer: 
 ->NORMAL
 ->write_pt [1392]
 ->read_pt [1530]
 ->size [2048 words]
 =>free [548 bytes]
 =>used [7636 bytes]
trace   Jun 23 00:44:20 func_onu���8X(36):2295333028: arrrg!
error   Nov 26 02:26:44 func_one9��8X(38):1644167343: oo+s� I�d-d it aga`n
ERROR: qb_rb_chunk_read failed: Connection timed out
[trace] ENTERING qb_rb_close()
[debug] Free'ing ringbuffer: /dev/shm/qb-create_from_file-header
```

jnpkrn:
This part definitely needs more love:

https://copr-be.cloud.fedoraproject.org/results/%40ClusterLabs/devel/epel-7-x86_64/00482520-libqb/build.log.gz
```
 ==== Corrupt blackbox test 14/32 ====
[debug] shm size:8192; real_size:8192; rb->word_size:2048
[debug] read total of: 8212
./blackbox-segfault.sh: line 17: 23805 Segmentation fault      (core dumped) ../tools/qb-blackbox butchered_blackbox.fdata
FAIL blackbox-segfault.sh (exit status: 1)
```

jnpkrn:
One of the ideas would be to add a cryptographic digest in addition
to mere checksum + perhaps add `--ignore-digest` and `--ignore-checksum`
switches that would allow to ignore the respective nonmatching value
at one's risk -- it would also be handy to have something like
`--ascii-only` that would substitute non-ascii characters in the deemed
output with a selected byte (e.g. `~` being the default).

sdake:
@jnpkrn a digital signature would be expensive, as the idea of the blackbox is to record all information coming from the system in realtime - even if the data doesn't quite make it to disk.  The solutions you mentioned for ascsi-ifying the text output would probably be helpful and some other guards in place around memory boundaries when printing log messages.

Cheers
-steve

jnpkrn:
For posterity, Python decides to stick with [SipHash](https://131002.net/siphash/)
[...]

PR #315, opened 2018-08-21
blackbox-segfault test - remove residual core files

grooverdan:
travis test failures on disttest showed corefiles left over. May as well remove them in the blackbox-segfault test.

### DIFF
--- a/tests/blackbox-segfault.sh
+++ b/tests/blackbox-segfault.sh
@@ -3,6 +3,7 @@
 # create a normal blackbox
 rm -f crash-test-dummy.fdata
 ./crash_test_dummy
+rm -f core*
 
 . ./test.conf
 


jnpkrn:
ok to test

fabbione:
ok to test

jnpkrn:
Ah, I remember now!
https://github.com/ClusterLabs/libqb/issues/237#issuecomment-263659152

Anyway, if the immediately preceding `qb-blackbox` run was what caused
the coredump at hand to be generated, your patch is not expected to
help since there's:
```
[ $? -gt 127 ] && exit 1 || true
```
and unless either the libtool's wrapper over the actual compiled binary
or the executing shell, assuming POSIX compatibility[*1], is botched,
the script will return earlier than where you put your clean command.

[*1]
http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_08_02:
> The exit status of a command that terminated because it received
> a signal shall be reported as greater than 128.

So the best solution appears to turn that AND-OR list into
properly conditionalized block where in the segfault case you
first run a command to print out the backtrace into stderr[*2],
then remove the core, the run `exit 1`.

[*2] see, e.g.,
https://github.com/jnpkrn/clufter/blob/v0.77.1/.travis.yml#L78-L82

jnpkrn:
Not to forget, thank you so much for pointing out this fishiness
with Travis CI!

jnpkrn:
Hmm, I start to see where this comes from.

Please, do the core removal right after
```
  3 # create a normal blackbox
  4 rm -f crash-test-dummy.fdata
  5 ./crash_test_dummy
```
where it belongs instead.

jnpkrn:
Pushed by hand (didn't trigger autoclose).
Thanks!

jnpkrn:
621b743207e0020ebd4e2aba0110c7b990f8c4dc

PR #318, opened 2018-08-27
Binutils 2.29 followup 2

jnpkrn:
Preferably goes after #288.


jnpkrn:
Should note when `PROTECTED` became also the possible outcome
(i.e., since which binutils version/commit).

jnpkrn:
Figured out the just the first commit was not disputable, the one
I already merged: 3af6104.

`tests: functional: make log_external accommodate line differences` one
is actually superceded with the more systemic counterpart in #288.

So I left one I am now pretty much doubting on the second thought,
which the previous comment meant to target (I figured out the missing
info and put into the commit, though).

Reason I am doubting it is not needed (and even could possibly be
harmful) is because Fedora rebuilds using both 2.29.1 and 2.30 seems
to be happy without such a change...

I guess I'll do some more testing so I can close this PR with a piece
in mind.

PR #320, opened 2018-09-04
Use git archive stored version tag

jfriesse:
Enhanced version of https://github.com/corosync/corosync/commit/0b0f32c56fb391dc6b093d7cb4aff4215eaa7f44 which should work with sed tag script. I'm not entirely happy with 3rd patch, because with little hacking it should be possible to reduce 3 cases into one call of git-version-gen, but this may be another PR.

### DIFF
--- a/.gitarchivever
+++ b/.gitarchivever
@@ -1,0 +1,1 @@
+ref names:$Format:%d$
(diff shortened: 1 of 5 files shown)


jnpkrn:
I take it that the purpose here is the on-the-fly archive consumption
from git in some CI pipelines.

Gave it some thoughts and came with a counter-proposal #321 where
my reasoning is captured as well.

Issue #312, opened 2018-07-05
libqb 1.0.3 PACKAGE_VERSION is UNKNOW


sasadangelo:
Hi,

I am trying to use libqb 1.0.3 with pacemaker and corosync to create a cluster and I think I found a bug. I compiled libqb as per doc but in /usr/lib/pkgconfig/libqb.pc in the Version field I found UNKNOWN.
This arise an issue in Pacemaker when I launch ./configure because it check with pkg-config if the library it is greater than 0.13. The check fails reporting that 0.13 was expected but UNKNOWN was found. I simply fixed the issue replacing the correct value in /usr/lib/pkgconfig/libqb.pc, but I think code must be fixed.

edwintorok:
Have you built libqb from the git tag or from the released tarball. I think the tarball differs from the git tag in that it has a .tarball-version inside it. 
Perhaps the right way of fixing this would be to enable `export-subst` in `.gitattributes` on the .pc file and use git archive substitution formats to get the version number substituted into the file.
That way when you generate a tarball from a git tag using `git archive` you would get the correct version inside it.

jnpkrn:
Sorry, but using "native on-the-fly archives" is not a supported
use case at this time.  Sadly, GitHub is obnoxious enough that we
cannot turn that "feature" (making for unintended misfeature here;
accidentally, it was discussed recently at the `developers` mailing
list that patchset review process leaves rather too much to be
desired, too).  In turn, we cannot be held accountable for something
we don't want to endorse in the first place, though we can probably
better document the matter.

Your options:

1. use proper (and stable and signed!) archives (you can recognize
   these for having a static file size that's displayed along the
   respective "Assets" per-release items)

2. `git clone [...]; git checkout $TAG; ./autogen.sh && ./configure [...]`
   possibly followed with, e.g., `make dist-xz`

Closing, though feel free to reopen if the above options don't suite
you for some reason.

jnpkrn:
![hopethishelps](https://user-images.githubusercontent.com/1391681/42352448-e1ca8e86-80ba-11e8-9899-40c28125154b.png)

chrissie-c:
Annoyingly github insists on putting the useless "Source Code" archives in the release list. I really wish they wouldn't.

sasadangelo:
Ok it worked. Thanks.

PR #321, opened 2018-09-04
build: allow for being consumed in a (discouraged) form of snapshots

jnpkrn:
This is meant as a lean, customized policy driven alternative
to the original proposal by Jan Friesse <jfriesse@redhat.com> that
balances the inherent trade-off in the opposite direction, so that the
configure.ac script is practically untouched and more weight and policy
is hardcoded in git-version-gen.  Problem with that approach stems from
(avoidable) effective fork of the respective gnulib's module and imposed
maintenance burden.

Speaking for libqb in particular, we should nonetheless make it
absolutely clear such in-development snapshots are nothing more,
nothing binding (not to think of viral injection of these bits
into circle of dependent packages upon their rebuilds), since the
changes accumulated since the last official release should only be
assumed firmly committed at the point the new release is cut (may
happen 99% of time with snapshots but no accountability from our
side for the complementary inter-release twists...), which is exactly
when many possibly unanticipated variables like correct SONAME
versions get to reflect what's appropriate.
Also, OpenPGP signature constitutes something more eligible for
one's trust than (provably) bit/content unstable archives without
the possibility of an independent authenticity/integrity verification.

  Therefore, the only thinkable and upstream-approved use cases
  for such snapshots are development-only purposes (CI et al.)!

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+/configure.ac export-subst
--- a/.gitignore
+++ b/.gitignore
@@ -16,24 +16,32 @@ autom4te.cache/
 config.status
 configure
 conftest.*
-.libs
 
-# since we use subdir-objects as automake option
-.dirstamp
+
+# dotfiles
+
+.*
+# - subsumed:
+#   .dirstamp (since we use subdir-objects as automake option)
+#   .libs
+#   .{,snapshot-,tarball-}version
+# - sans:
+!.git*
+!/.tito
+!/.travis.yml
+
 
 # ignore "libtoolized" m4 files, but keep our (custom-prefixed) ones
 /m4/*
 !/m4/ax_*.m4
 
 # build-aux/release.mk related litter
-/.tarball-version
 /tag-*
 
 /lib/qblog_script.ld
 # already captured with wildcard: /lib/qblog_script.la
 
 libtool
-.version
 libqb.spec
 libqb-*
 cov
--- a/Makefile.am
+++ b/Makefile.am
@@ -48,7 +48,7 @@ doxygen:
 	$(MAKE) -C docs doxygen
 
 dist-clean-local:
-	rm -f autoconf automake autoheader
+	rm -f .snapshot-version autoconf automake autoheader
 
 # this will also get rid of "libtoolized" m4 files
 maintainer-clean-local:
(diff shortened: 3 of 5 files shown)


chrissie-c:
The travis CI build failure looks odd for this. is it related or is it just Travis being weird (again)?

jnpkrn:
Looks like unanticipated, activistic "seriously? you might cause something!",
rather unwanted "thoughtfulness". Let me try something with `umask` or `chmod`.

jnpkrn:
Of course, it was a matter of `make distcheck` that copies sources
into a separate directory marking them read-only (so as to catch the
case that the build recipes attempt to modify source files at their
source location, which is an error), and consequently, such untouched
sources to be distributed are copied to distribution directory
(non-writability being retained), but there, it's legal to modify
files from dist-hook, but as documented, one should rather mark
them writable again first (and `mv` is actually reasonably safe
in a way it asks for confirmation when read-only file is to be
replaced, which is exactly what made the CI timeout elapse).

Now the file gets marked as writable prior to being replaced
(as opposed to `mv -f` since we'd like to know about any other
possible issues).

Also, noticed that `autoreconf` needs to be invoked with `-f` when
the `VERSION` in generated `configure` script needs to be reflected
truthfully (another precondition is that no .*version files are
present, i.e., at least `make maintainer-clean`, if not `distclean`,
shall be called in advance, but that shall be clear already).
Hence extended `autogen.sh` to that effect.

jfriesse:
Just a few (biased) comments:
- If you try diff gnulib and libqb version of `git-version-gen` you will find that they already differs. So technically speaking, you are already maintain a fork
- #320 is based on assumption, that non version tagged archives should not be built at all. This PR is based on idea to "care to bump "X.Y.Z-yank" template below upon each release very desirable". Isn't that going against main idea of git-version-gen?
- Eventho I must say I quite like this alternative aproach, it is showing it's weakness because configure.ac has to be modified to pristine version during dist-hook.
- As long as I didn't overlooked something, make rpm will not work and alternative to https://github.com/ClusterLabs/libqb/pull/320/commits/0b07907eba509a483d1e6f4dc08eecddf8fa078b should be ported.

chrissie-c:
I'm generally in favour of this approach. The language in the messages needs updating to be a little less ... archaic though.

jnpkrn:
On 11/09/18 00:31 -0700, Chrissie Caulfield wrote:
> I'm generally in favour of this approach. The language in the
> messages needs updating to be a little less ... archaic though.

The purpose to fit those messages into 80 characters was fulfilled
and this is turning into bikeshedding around exact wording for
a marginal part nobody but the developers should ever see,
I am afraid (note that the genuine tarballs come with configure
script readily pregenerated).

-- 
Jan (Poki)

jfriesse:
@jnpkrn I don't understand why are you still talking something about CI. The reason for implementing that feature was (at least for corosync) to mitigate problems with github release page, where people tends to use "Source code" link (probably not that much problem for libqb, but quite a big for projects which doesn't use release page at all (corosync/kronosnet)). If it also helps CI, good, but it's (at least for me) just nice side effect.

jnpkrn:
On 11/09/18 02:52 -0700, Jan Friesse wrote:
> @jnpkrn I don't understand why are you still talking something about
> CI. The reason for implementing that feature was (at least for
> corosync) to mitigate problems with github release page, where
> people tends to use "Source code" link (probably not that much
> problem for libqb, but quite a big for projects which doesn't use
> release page at all (corosync/kronosnet)). If it also helps CI,
> good, but it's (at least for me) just nice side effect.

Frankly, I don't care about corosync/kronosnet context here.
The choice has already been done there and it may work for
these projects and their workflows well, while the same would
be suboptimal for libqb.

In libqb context, we have proper releases and that logic by GitHub
damages that intention without a possibility of turning this redundant
"well-intentionessness" off (is this accidental or on purpose?
for instance, GitLab seems to be a lot less agressive about leading
bycomers to download the per-tag snapshots:
https://gitlab.gnome.org/GNOME/vte/tags).

I think hosting these proper releases at clusterlab.org was discussed
at some point in the past, perhaps it would be the most sane thing to
do, along with instructing people in README where to look for new
releases.

Consequently, if the consumability backed with "git archive"
is desired for libqb regardless, the only thinkable use-case is CI
and development in general.  When unfortunately confused people get
to download such a form also for tagged release, they'd get warned
by `./autogen.sh` they likely don't want to use that as opposed
to proper releases (but at least, if they are insistent, they'll
avoid `UNKNOWN` version indication now as well, see #312).
I think that's preferred and the most sane outcome we can get.

A side note since I've been looking at associated issues: one
cannot use `git archive` locally with GH location directly, it
seems: https://github.com/isaacs/github/issues/554
Perhaps one more incentive to move away from GH as an authoritative
code hosting, along with the mentioned overly aggressive emphasis on
"get tarball from us [and don't even think about looking elsewhere]"
(https://lists.clusterlabs.org/pipermail/developers/2018-June/001241.html)

-- 
Jan (Poki)

PR #324, opened 2018-09-20
UPDATED: doc (ABI comparison) and various other fixes

chrissie-c:
This is an updated version on Poki's PR#287. It fixes 3 things
1. Updated to latest tests/Makefile.am
2. Compatibility with check 0.9 (uints were introduced in 0.10)
3. Fixed C++ compiler check

4aa9d08 can be ignored as it's reverted by cf6c346. I'll squash them on commit.


tests/print_ver.c
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016 Red Hat, Inc.
+ * Copyright 2018 Red Hat, Inc.
```
jfriesse: I do not want to contribute into meaningless discussion, but this really should be `Copyright (C) 2016-2018`, as an example see `/usr/include/*.h` (for example `math.h`)

tests/check_list.c
```diff
@@ -0,0 +1,107 @@
+/*
+ * Copyright 2018 Red Hat, Inc.
```
jfriesse: (C) missing

tests/print_ver.c
kgaillot: Almost right :) It could be either Copyright 2016,2018 or Copyright 2016-2018. The (C) convention, while common, does not conform to copyright regulations.

chrissie-c:
I think I'd rather go for consistency and leave the (c) in - thanks for spotting that. AFAIR when we asked for legal advice on this, the response what that it actually doesn't much matter what we put in that line from their point of view.

jfriesse:
ACK by me if "(c)" (year one) issue is fixed.

tests/print_ver.c
jfriesse: Ok, thank you a lot for explanation (and confirmation that law and common sense are not always in agreement :) )

kgaillot: Oh, and you are correct that the original year is the most important one to put in the copyright notice. The purpose of a copyright notice is to establish a claim to priority (in the sense of "which came prior") and the date at which copyright protections expire.

Copyright has long had to deal with publications that are updated, such as encyclopedias, so there is some established usage there, though it is not specified by law: either a range like 2014-2018 or a list of specific years like 2014,2015,2018.

https://techwhirl.com/updating-copyright-notices/

The legal side hasn't kept up with the idea of projects updated by many contributing copyright holders, so the law does not specify a format for that, and it is only convention. I've seen multiple variations; some list only the original copyright holder, some put a separate notice for modifications by a later author, some put the original holder "and the <project> contributors".

It's primarily an intellectual exercise though, as the chances of someone bringing legal action for copyright issues in libqb is vanishingly small, and the chance that the copyright notice line would have any significant bearing on that is even more infinitesimal.

PR #326, opened 2018-10-01
log: Add option to re-open a log file

chrissie-c:
With an optional new filename

### DIFF
--- a/lib/log_file.c
+++ b/lib/log_file.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2011 Red Hat, Inc.
+ * Copyright (C) 2011-2018 Red Hat, Inc.
  *
  * All rights reserved.
  *
@@ -60,20 +60,50 @@ _file_close(int32_t t)
 	FILE *f = qb_log_target_user_data_get(t);
 
 	if (f) {
-		fclose(f);
 		(void)qb_log_target_user_data_set(t, NULL);
+		fclose(f);
 	}
 }
 
+static int
+_do_file_reload(const char *filename, int32_t target)
+{
+	struct qb_log_target *t = qb_log_target_get(target);
+	FILE *oldfile = qb_log_target_user_data_get(target);
+	FILE *newfile;
+	int rc = 0;
+
+	if (filename == NULL) {
+		filename = t->filename;
+	}
+	newfile = fopen(filename, "a+");
+
+	qb_log_thread_pause(t);
+
+	if (newfile) {
+		/* Only close oldfile if newfile open succeeds */
+		if (oldfile) {
+			fclose(oldfile);
+		}
+
+		if (filename != t->filename) {
+			(void)strlcpy(t->filename, filename, PATH_MAX);
+		}
+		(void)qb_log_target_user_data_set(target, newfile);
+	}
+	else {
+		rc = -1;
+	}
+	qb_log_thread_resume(t);
+
+	return rc;
+}
+
+/* The version called from the logger object */
 static void
 _file_reload(int32_t target)
 {
-	struct qb_log_target *t = qb_log_target_get(target);
-
-	if (t->instance) {
-		fclose(t->instance);
-	}
-	t->instance = fopen(t->filename, "a+");
+	(void)_do_file_reload(NULL, target);
 }
 
 int32_t
@@ -123,4 +153,18 @@ void
 qb_log_file_close(int32_t t)
 {
 	qb_log_custom_close(t);
+}
+
+int32_t
+qb_log_file_reopen(int32_t t, const char *filename)
+{
+	int rc;
+
+	rc = _do_file_reload(filename, t);
+
+	if (rc) {
+		return -errno;
+	} else {
+		return 0;
+	}
 }
--- a/lib/log_file.c
+++ b/lib/log_file.c
@@ -71,12 +71,14 @@ _do_file_reload(const char *filename, int32_t target)
 	struct qb_log_target *t = qb_log_target_get(target);
 	FILE *oldfile = qb_log_target_user_data_get(target);
 	FILE *newfile;
-	int rc = 0;
+	int saved_errno;
+	int rc;
 
 	if (filename == NULL) {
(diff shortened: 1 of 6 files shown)


chrissie-c:
This changes the API of the custom loggers (to take a struct timespec rather than a time_t) but we're already changing the soname of the library anyway for v2.0.0

wferi:
> This changes the API of the custom loggers (to take a struct timespec rather than a time_t)

How does time representation influence log reopening? Thanks.

chrissie-c:
> > This changes the API of the custom loggers (to take a struct timespec rather than a time_t)
> 
> How does time representation influence log reopening? Thanks.

It doesn't - this is a different PR I'm still working on, sorry. Blame it on Mondays!

lib/log_file.c
```diff
@@ -69,11 +69,14 @@ static void
...
+	FILE *oldfile = qb_log_target_user_data_get(target);
+	FILE *newfile;
 
-	if (t->instance) {
-		fclose(t->instance);
+	if (oldfile) {
+		fclose(oldfile);
 	}
-	t->instance = fopen(t->filename, "a+");
+	newfile = fopen(t->filename, "a+");
```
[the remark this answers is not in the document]
chrissie-c: Good points all. I've pushed a new commit to fix them

lib/log_file.c
```diff
@@ -132,13 +148,14 @@ int32_t
...
 		(void)strlcpy(target->filename, filename, PATH_MAX);
 	}
-	_file_reload(t);
+	rc = _do_file_reload(t);
 
-	if (!target->instance) {
+	if (rc) {
 		return -errno;
 	} else {
 		return 0;
```
wferi: Why not move this part into `_do_file_reload()` so that you can simply `return _do_file_reload(t)` here? Gets rid of a conditional and moves reading of `errno` much closer to the function setting it, which adds to clarity IMO.

lib/log_file.c
```diff
@@ -132,13 +148,14 @@ int32_t
 qb_log_file_reopen(int32_t t, const char *filename)
 {
 	struct qb_log_target *target = qb_log_target_get(t);
+	int rc;
 
 	if (filename) {
 		(void)strlcpy(target->filename, filename, PATH_MAX);
```
wferi: Isn't changing the filename field premature here? If `open()` fails, we keep the old `FILE` structure after all. I overlooked this added complication; looks like one must pass the new file name to `_do_file_reload()` in a separate argument...

lib/log_file.c
chrissie-c: We can't do that because _do_file_reload() is called from the logger object and that doesn't return an error (bad design, I know, not guilty). That's why I had to split it into two functions in the first place!

wferi: You mean `_file_reload()`, don't you? You could still define it to `errno = -_do_file_reload(t, NULL)` I guess (using the second argument for passing in the new filename as discussed in my other comment). Then get rid of it altogether with the next API change.

lib/log_file.c
wferi: I believe you. It might be useful to close this avenue of future confusion anyway.

chrissie-c:
I've changed the reopen to only close the old file if the new one succeeds, and moved the fclose to be sure that errno is retained on return.

Return code from _file_reload() I'm going to leave as-is for the moment. Changing the logger api is a bigger job and should be kept al together IMO.

lib/log_file.c
```diff
@@ -147,13 +153,9 @@ qb_log_file_close(int32_t t)
...
 qb_log_file_reopen(int32_t t, const char *filename)
 {
-	struct qb_log_target *target = qb_log_target_get(t);
 	int rc;
 
-	if (filename) {
-		(void)strlcpy(target->filename, filename, PATH_MAX);
-	}
-	rc = _do_file_reload(t);
+	rc = _do_file_reload(filename, t);
```
wferi: Why not `return -errno; /* from fopen() */` instead of `return -1;` in `_do_file_reload()` so that you can then rename it to `int32_t qb_log_file_reopen()` and happily be one function down?

lib/log_file.c
```diff
@@ -60,20 +60,45 @@ _file_close(int32_t t)
...
+	if (filename == NULL) {
+		filename = t->filename;
+	}
+	newfile = fopen(filename, "a+");
 
-	if (t->instance) {
-		fclose(t->instance);
+	/* Only close oldfile if newfile succeeds */
+	if (oldfile && newfile) {
+		fclose(oldfile);
```
wferi: I'd write this as `if (oldfile) { fclose(oldfile); }` inside the `if (newfile) { }` block below for clarity. YMMV.

wferi:
Wouldn't dropping support for changing the log file name during reopen help a little? That's somewhat unrelated anyway and remains possible by stopping and starting logging the old way.

jfriesse:
@wferi Not too much. Current logging thread is simply not protected from changing configuration at all (so qb_log_ctl2) and errors are visible even when something as simple as reload corosync.conf is performed (it basically works just by accident).

chrissie-c:
To clarify what's happening here. I force pushed a new patch yesterday that adds locking around all of the logging reconfiguration. Today's small amendment is to make 'in_logger' atomic as it's accessed across threads and helgrind was complaining (it still complains but only because it only understands pthreads).

jfriesse:
Last update seems to solve all problems with `corosync-cfgtool -R` and logging , so ACK by me.

Looks like some of @wferi's comments are still unresolved (Why not return -errno; ...) so probably not yet ready to go.

chrissie-c:
Thanks. I'd rather leave it as it is TBH, then the two 'API' functions are equal and the fn that does the work is called by both.

wferi:
Your call, certainly. I probably wouldn't have brought it up if `_do_file_reload()` wasn't introduced in this very pull request. If you prefer to have this symmetry between `_file_reload()` and `qb_log_file_reopen()`, then be it.

chrissie-c:
Thanks.

PR #327, opened 2018-10-02
log: Add configure-time option to use systemd journal instead of syslog

chrissie-c:
As most of this patch is in the config system I'm hoping some nice person will fix it up for me - it does (seem to ) work, but I suspect it's not optimal.

The actual code is trivial.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -149,6 +149,17 @@ AC_SEARCH_LIBS([clock_gettime], [rt],,[AC_MSG_ERROR([cannot find clock_gettime()
 AC_SUBST([rt_LIBS],[$LIBS])
 AX_RESTORE_FLAGS
 
+USE_JOURNAL="no"
+PKG_CHECK_MODULES([SYSTEMD], [libsystemd], [have_systemd=yes],[have_systemd=no])
+if test "x$enable_systemd_journal" = "xyes" ; then
+  if test "x$have_systemd" = "xyes" ; then
+     AC_DEFINE_UNQUOTED(USE_JOURNAL, 1, [Use systemd journal logging])
+     USE_JOURNAL="yes"
+  else
+     echo "systemd libraries not found, will just use syslog"
+  fi
+fi
+
 # look for testing harness "check"
 PKG_CHECK_MODULES([CHECK], [check >= 0.9.4],[with_check=yes],[with_check=no])
 AM_CONDITIONAL(HAVE_CHECK, test "${with_check}" = "yes")
@@ -538,6 +549,9 @@ AC_ARG_WITH([socket-dir],
 	[ SOCKETDIR="$withval" ],
 	[ SOCKETDIR="$localstatedir/run" ])
 
+AC_ARG_ENABLE([systemd-journal],
+  [AS_HELP_STRING([--enable-systemd-journal],[Allow use of systemd journal instead of syslog])])
+
 AC_ARG_WITH([force-sockets-config-file],
   [AS_HELP_STRING([--with-force-sockets-config-file=FILE],[config file to force IPC to use filesystem sockets (Linux & Cygwin only) @<:@SYSCONFDIR/libqb/force-filesystem-sockets@:>@])],
 	[ FORCESOCKETSFILE="$withval" ],
@@ -701,7 +715,6 @@ AC_SUBST([FORCESOCKETSFILE])
 AC_DEFINE_UNQUOTED([SOCKETDIR], "$(eval echo ${SOCKETDIR})", [Socket directory])
 AC_DEFINE_UNQUOTED([LOCALSTATEDIR], "$(eval echo ${localstatedir})", [localstate directory])
 AC_DEFINE_UNQUOTED([PACKAGE_FEATURES], "${PACKAGE_FEATURES}", [quarterback built-in features])
-
 AC_DEFINE_UNQUOTED([FORCESOCKETSFILE], "$(eval echo ${FORCESOCKETSFILE})", [for sockets config file])
 
 # version parsing (for qbconfig.h)
@@ -763,6 +776,7 @@ AC_MSG_RESULT([  State information        = ${localstatedir}])
 AC_MSG_RESULT([  System configuration     = ${sysconfdir}])
 AC_MSG_RESULT([  SOCKETDIR                = ${SOCKETDIR}])
 AC_MSG_RESULT([  Features                 = ${PACKAGE_FEATURES}])
+AC_MSG_RESULT([  Use systemd journal      = ${USE_JOURNAL}])
 AC_MSG_RESULT([])
 AC_MSG_RESULT([$PACKAGE build info:])
 AC_MSG_RESULT([  Optimization             = ${OPT_CFLAGS}])
--- a/lib/log.c
+++ b/lib/log.c
@@ -1135,11 +1135,15 @@ qb_log_ctl2(int32_t t, enum qb_log_conf c, qb_log_ctl2_arg_t arg_not4directuse)
 		conf[t].ellipsis = arg_i32;
 		break;
 	case QB_LOG_CONF_USE_JOURNAL:
+#ifdef USE_JOURNAL
 		if (t == QB_LOG_SYSLOG) {
 			conf[t].use_journal = arg_i32;
 		} else {
 			rc = -EINVAL;
 		}
+#else
+		rc = -EOPNOTSUPP;
+#endif
 		break;
 
 	default:
--- a/lib/log.c
+++ b/lib/log.c
@@ -1134,6 +1134,13 @@ qb_log_ctl2(int32_t t, enum qb_log_conf c, qb_log_ctl2_arg_t arg_not4directuse)
 	case QB_LOG_CONF_ELLIPSIS:
 		conf[t].ellipsis = arg_i32;
 		break;
+	case QB_LOG_CONF_USE_JOURNAL:
+		if (t == QB_LOG_SYSLOG) {
+			conf[t].use_journal = arg_i32;
+		} else {
+			rc = -EINVAL;
+		}
+		break;
 
 	default:
 		rc = -EINVAL;
--- a/tests/check_log.c
+++ b/tests/check_log.c
@@ -1027,6 +1027,8 @@ END_TEST
 #else
 START_TEST(test_syslog)
 {
+	int rc;
+
 	qb_log_init("flip", LOG_USER, LOG_INFO);
 	qb_log_ctl(QB_LOG_SYSLOG, QB_LOG_CONF_ENABLED, QB_TRUE);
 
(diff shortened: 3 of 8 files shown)


configure.ac
```diff
@@ -145,6 +146,13 @@ AC_SEARCH_LIBS([gethostbyname], [nsl],,[AC_MSG_ERROR([cannot find gethostbyname(
 AC_SUBST([nsl_LIBS],[$LIBS])
 AX_RESTORE_FLAGS
 
+if test "x$enable_systemd_journal" = "xyes" ; then
+  AX_SAVE_FLAGS
+  AC_SEARCH_LIBS([sd_journal_send], [systemd-journal],,[AC_MSG_ERROR([cannot find sd_journal_send() function])])
```
jfriesse: I would recommend using `PKG_CHECK_MODULES` (because it sets correctly not only the LIBS but also the CFLAGS.

Also (without deep knowledge about history) this code seems to work for CentOS 7, but Fedora 28+ seems not to have systemd-journal library at all and instead it's using libsystemd (https://www.freedesktop.org/software/systemd/man/sd-journal.html)

configure.ac
```diff
@@ -146,13 +146,19 @@ AC_SEARCH_LIBS([gethostbyname], [nsl],,[AC_MSG_ERROR([cannot find gethostbyname(
...
+PKG_CHECK_MODULES([SYSTEMD], [libsystemd], [have_systemd=yes],[have_systemd=no])
 if test "x$enable_systemd_journal" = "xyes" ; then
-  AX_SAVE_FLAGS
-  AC_SEARCH_LIBS([sd_journal_send], [systemd-journal],,[AC_MSG_ERROR([cannot find sd_journal_send() function])])
-  AC_SUBST([systemd_LIBS], [$LIBS])
-  AX_RESTORE_FLAGS
+  if test "x$have_systemd" = "xyes" ; then
+     AC_DEFINE_UNQUOTED(USE_JOURNAL, 1, [Use systemd journal logging])
+     USE_JOURNAL="yes"
+     echo "CC: found systemd & journal requested"
```
jfriesse: This debug print should go or be replaced by something without CC: prefix :)

examples/Makefile.am
```diff
@@ -25,27 +25,27 @@ noinst_PROGRAMS = mapnotify simplelog tcpclient tcpserver ipcclient ipcserver
 
 mapnotify_SOURCES = mapnotify.c $(top_builddir)/include/qb/qbmap.h
 mapnotify_CPPFLAGS = -I$(top_builddir)/include -I$(top_srcdir)/include
-mapnotify_LDADD = $(top_builddir)/lib/libqb.la
+mapnotify_LDADD = $(top_builddir)/lib/libqb.la $(SYSTEMD_LIBS)
```
jfriesse: I don't think this adding $(SYSTEMD_LIBS) everywhere is good idea (and if it really is, then ` libqb.pc` must be changed).

What about changing lib/Makefile.am` and add $(SYSTEMD_CFLAGS) and $(SYSTEMD_LIBS) to libqb_la_CFLAGS and libqb_la_LIBADD? Then libqb.la gets dependency on systemd (when needed) and apps using libqb will need no change? (please correct me if I'm wrong, I'm not libtool magic master).

configure.ac
chrissie-c: err yes. That should never have been left in. TBH I only pushed it to move a copy to my Fedora system for testing!

examples/Makefile.am
chrissie-c: Possibly. I'm struggling with auto-stuff  too, It will definitely need to go in the pkgfile somehow

configure.ac
```diff
@@ -531,6 +546,9 @@ AC_ARG_WITH([socket-dir],
 	[ SOCKETDIR="$withval" ],
 	[ SOCKETDIR="$localstatedir/run" ])
 
+AC_ARG_ENABLE([systemd-journal],
+  [AS_HELP_STRING([--with-systemd-journal=DIR],[Use systemd journal instead of syslog])])
```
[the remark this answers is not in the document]
chrissie-c: oops, copy/paste laziness.

chrissie-c:
Thanks for that - it never occurred to me that both might be available. Not being a Debian user. Do you have any suggestions on how it might be made switchable? 

Having said that, it would need the systemd libs to be dragged in even on a non-systemd installation if it was made run-time switchable - which I suspect would be less than popular given some of the opinions on systemd! In such circumstances it might just be easier to build for syslog and leave it at that. It will still work on systemd systems but you wouldn't get the logging metadata of course?

I'm interested in opinions here.

kgaillot:
Maybe overkill, but it would be possible to add build-time support for run-time selection. I.e. a configure option for system log that could be set to syslog, systemd, or both, and if both, that would enable a run-time option to select syslog or systemd. Distros could pick one option or even offer multiple variants to get around the dependency issue. (I hadn't thought about that before, but a distro like Debian could even offer a pacemaker-systemd package that depended on libqb-systemd, with systemd support enabled in both, and non-systemd packages without that support.)

wferi:
Cleanest would be to make it an API choice in addition to file and syslog. The obvious drawback is that all applications would have to parse and act on an additional config option, so you get `to_file`, `to_syslog` and `to_journal` for Corosync. The ugly solution is to do a runtime selection based on an environment variable. The additional small library dependency is not a problem.

wferi:
That would be perfectly fine from the library packager's point of view (let's hope I don't forget to bump the symbol version info for this ABI extension). Yeah, it's a little awkward to pull in a new library for a single function, but requiring applications to pass in the address of `sd_journal_send()` would be even more so...
Please note that the current name of the unified library is libsystemd, with oldish systems requiring a fallback to libsystemd-journal.

jfriesse:
@wferi Current code is already using unified libsystemd. IMHO we don't need to try to keep backwards compatibility with libsystemd-journal (I found it in Debian wheezy (already unsupported), jessie (only for some architectures), and in RHEL library still exists but only to keep "compatibility" and main library is libststemd (since 7.2 which is already EOL)).

wferi:
@jfriesse Even better. I remembered the original patch using the legacy library, and Chrissie mentioned it again in her latest comment, so I decided to mention it again. I'm all for forgetting about the old name if it's acceptable for you.

chrissie-c:
it was just a slip of the keyboard really, partly also cause by the fact that my desktop is running RHEL7 ;)

chrissie-c:
I've force committed an updated patch that should have tidied everything up. see what you think

wferi:
All those conditionals in the `log_syslog.c` functions aren't too pretty; what's the reason for not introducing a new `QB_LOG_JOURNAL` static target slot? You could define the single needed method in a `log_journal.c` conditional module without further compile- and runtime conditionals (apart from the initialization stuff). Maybe not very useful, but this would also enable using syslog and journal loggers concurrently.

tests/check_log.c
```diff
@@ -966,7 +1016,15 @@ log_suite(void)
 #endif
 	add_tcase(s, tc, test_extended_information);
 	add_tcase(s, tc, test_zero_tags);
+/*
+ * You can still use syslog and journal in a normal application,
+ * but the syslog_override code doesn't work when -lsystemd is present
+ */
```
wferi: Is this some load order problem?

chrissie-c:
because that's what seemed to be the consensus in earlier discussions?

chrissie-c:
But also, I don't see the point in having both available at the same time. If systemd is installed then sending to syslog AND the journal is sending the message to the same place twice, it's just that one has the metadata and one doesn't.

tests/check_log.c
chrissie-c: I tried moving the link order around with no luck. TBH I don't care enough to spend time on it as it's only a test program. If you do, then I'm happy to take patches ;-)

wferi:
I'm sorry for the confusion, it's entirely my fault: I overestimated the scope of `qb_log_ctl()`. In my opinion the implementation is crying for a new target type, both for efficiency and cleanliness, but I acknowledge that I may miss several points and our tastes may well differ as well. You're also right that having both active would send the messages to the same place twice in 100% of the actual configurations, though that's a question of system configuration (policy), and keeping policy out of libraries is considered good design. And there's also that remaining 0% of setups..:) Anyway, my main motivation for suggesting the change was code cleanliness first, efficiency second and flexibility fifth.

chrissie-c:
No problem. I don't have a strong opinion on which way to add this in. My very first thought was to do as you suggest and add a whole new target type but it seemed overkill.

kgaillot:
I like sharing the slot. I see the systemd journal more as an enhanced syslog variant than an independent mechanism, most obviously because it takes over the logger command and syslog() function, but also because it serves the same logical function (aggregating log output from daemons). It has a separate API if someone wants to use it, but that's on top of the traditional syslog API (i.e. more like an extension).

jfriesse:
I'm also for qb_ctl mechanism instead of new target type. AFAIK when systemd is used as init system then journald is used and it owns /dev/log without any way to disable such behavior. Then syslog = journald. Contrary on systems without systemd, journald is not used and only syslog make sense. So it's basically on/off switch, not binary none, one or other or both.

lib/log.c
```diff
@@ -1127,6 +1127,13 @@ qb_log_ctl2(int32_t t, enum qb_log_conf c, qb_log_ctl2_arg_t arg_not4directuse)
 	case QB_LOG_CONF_ELLIPSIS:
 		conf[t].ellipsis = arg_i32;
 		break;
+	case QB_LOG_CONF_USE_JOURNAL:
+		if (t == QB_LOG_SYSLOG) {
+			conf[t].use_journal = arg_i32;
+		} else {
+			rc = -EINVAL;
+		}
+		break;
```
wferi: The current code silently ignores this operation if `USE_JOURNAL` is not defined. Wouldn't it be better to return some error instead?

wferi:
The above arguments for disallowing parallel use hinge on `/dev/log` being a symlink to `/run/systemd/journal/dev-log`. This is arranged by the stock `systemd-journald-dev-log.socket` unit, so it's a strong standard, but not something entirely unavoidable.
But anyway, this does not change that from a higher, "semantic" point of view these arguments are sound, and I've got no reason to challenge them.
What bugs me is the ugly code for making `t->close()` and `t->reload()` no-ops instead of setting the ops pointers to `NULL`. The cost of that extra `if` in `_syslog_logger()` is pretty much amortized by the priority, buffering and formatting code (which could be factored out), but still I think getting rid of the `use_journal` flag and changing the ops pointers instead would result in cleaner code. I now realized that it does not matter whether journal logging is a new target type or a target option in this regard.

kgaillot:
@jfriesse I recently had the same issue with journald rate-limiting messages from CTS runs; this was my fix:

    mkdir -p /etc/systemd/journald.conf.d
    echo RateLimitIntervalSec=2s >> /etc/systemd/journald.conf.d/cts.conf

By default it rate-limits after 10,000 messages in 30 seconds from a single package. That changes it to 10,000 messages in 2 seconds. (It's possible to disable it altogether, but I read that that causes performance issues.)

lib/log.c
chrissie-c: Good point. commit added (with test)

chrissie-c:
Thanks for all the helpful comments on this.

PR #330, opened 2018-11-07
[build] fix supported compiler warning detection

fabbione:
move from AC_PREPROC_IFELSE (strongly discouraged) to AC_COMPILE_IFELSE

our detection system was very weak and recent versions of clang did
show that PREPROC_IFELFE (cpp) would enable warning options that
the compiler does not support (clang).

use a full compilation test to detect what works and what doesn't.

Also expand the warning list to include new / renamed clang options
of equivalents already enabled for older versions of clang and gcc.


chrissie-c:
It doesn't seem to add any extra warnings that I can see on gcc 4,8 or clang 5. 

Thanks.

PR #331, opened 2018-11-09
log: Remove more dead code from linker callsites

chrissie-c:
also make some of the comments more readable.


chrissie-c:
More to come on this ....

jnpkrn:
Yes, I am just crying how self-proclaimed powers could, without
carring to even understand which problems were being solved(!!!),
destroy one of two incentives to use libqb over other logging
solutions out there, while patting on their backs when doing so.

Yay, it was really your master piece, I'll tell you

PR #332, opened 2018-12-09
Add Pthreads (and possibly other) flags to the pkg-config file

wferi:


jnpkrn:
Just gathering the context, is this to serve building libqb
(it's users?) statically or does just `Libs.privates` receive
a special treatment in Debian or something like that?

jnpkrn:
Thanks, was just curious since I vaguely remember `Libs.private`
seldom references scattered in the Fedora devel ML and casting
some opinionated ideas about its proper treatment
\+ there's at least a single crucial difference regarding libtool
already as discussed elsewhere under this project, for instance.

Can you add this "static compiling" + "completeness" justifications
to the commit message, please?  There's some bad experience with
not entirely obvious changes that charge an extra time to understand
in retrospect if need be, so better to prevent that systemically

Also would mention that gethostbyname/libnsl counterpart doesn't
qualify since it's only used in the examples (if I'm not missing
anything), not in the library proper -- in case one would wonder
in the plain cross-check.

Thanks

wferi:
Thanks for the merge.
Why/how was my full name replaced in the commit? I don't mind, just curious...

chrissie-c:
I'm not sure. github being weird I suspect!  I should maybe do manual commits if it's going to start happening regularly. 

which removes half the reason for having GH features in the first place. *sigh*

Sorry

jnpkrn:
Happens easily when single-button-laziness prevails, and I've already
pointed this out in the past (over a year ago if not earlier):

https://github.com/ClusterLabs/libqb/pull/251#issuecomment-302414542

So yes, please finally do retain a full control over the commits to hit
the tree, everything else carries this spoiling risks (and that GitHub
will continue to be the primary forge is far from being set in stone,
especially when none of us has a bit of a power over the hosting).

That would also avoid the silliness of undesired authorship handling:
https://github.com/ClusterLabs/libqb/commit/44a9379c89e2fd0a2d09df79e2a3eaf1d1b4a5c7

Hopefully that's enough of a justification.

PR #333, opened 2018-12-11
skiplist: Fix previous skiplist fix

chrissie-c:
The last fix to skiplist never ran the code that patched up the level
list as it updated the current level before running the loop.

This now works.

### DIFF
--- a/lib/skiplist.c
+++ b/lib/skiplist.c
@@ -505,10 +505,11 @@ skiplist_rm(struct qb_map *map, const char *key)
 	   (header) preceding @found_node, which can be distinguished with
 	   NULL being used as a key (second allowing condition below). */
 	if (found_node->refcount > 1 || cur_node->key == NULL) {
-		found_node->level = SKIPLIST_LEVEL_MIN - 1;  /* no "forward" drop */
+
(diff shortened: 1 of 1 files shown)


jnpkrn:
Ack, btw. that's what review is meant to capture...

Looking back at #302 I think the concern with `test_skiplist_notifications`,
it was either fixed directly within the temporally later commit
(if the presentation within UI is to be trusted, GitHub used to be
silly about forgetting older iterations of PR, luckily not any more)
or it was regarding whether the notifiers access either of faked
fields (`level`, `forward`), which is not immediately possible.

There are some doubts, though, regarding having 2+ iterators throughout
the "map" and deciding to destroy it all at the particular "peculiar"
moment.

jnpkrn:
> There are some doubts, though, regarding having 2+ iterators throughout
> the "map" and deciding to destroy it all at the particular "peculiar"
> moment.

Indeed it looks bad, but that's outside of this scope.

Will just drop the extraneous line that was kept around and merge it.

jnpkrn:
Re `merges:`: https://github.com/isaacs/github/issues/548

Issue #222, opened 2016-09-12
Using /dev/shm directly is (potentially) broken


jnpkrn:
See https://wiki.debian.org/ReleaseGoals/RunDirectory#Packages_using_.2Fdev.2Fshm

chrissie-c:
Good catch, I'll look into this.

jnpkrn:
Few notes:
- using `/var/run` (aka `SOCKETDIR` on standard Linux `./configure` pass) for a shared memory mapped files location (further: `shmdir`) would not be very fortunate as this is still disk-backed by default with some (especially older) systems, unlike relatively new `/run` with clear (matching) intentions, which should be preferred
- there's this long-upcoming change to autoconf that would standardize `--runstatedir` option to `./configure`:
 http://git.savannah.gnu.org/cgit/autoconf.git/commit/?id=a197431414088a417b407b9b20583b2e8f7363bd (something to keep in mind for future, like being a default for `SOCKETDIR`)
- `./configure` script could have an option for setting `shmdir` explicitly, otherwise it would default to what `SOCKETDIR` uses, unless it would start with `/var/run` in which case it would fallback to `/dev/shm` on Linux/Cygwin (as is currently the case)
- last but definitely not least (quite the contrary!), library should allow the shared memory backed server sides to specify full path for the handshake files per their choosing if they opt-in to (e.g., so as to utilize `/run/<program>/' hierarchy which is encouraged in systemd-enabled environments)

wferi:
Hi!
Thinking about filesystem path choices, I reached a couple of questions:
* The abstract sockets are Linux-specific, and you want to support the portable filesystem sockets as well, so why keep the abstract socket support at all?
* Abstract sockets enable using libqb in unprivileged applications on Linux. Why not set the default socket directory to `/tmp` (or even create a temporary directory an use that) to enable unprivileged use on BSDs as well?
* Using `mmap()` on files under `/dev/shm` is pretty much what the POSIX shared memory API does. Why not use the standard shared memory API and forget about system specific paths?

What do you think?
Regards,
Feri.

chrissie-c:
In general I'm very much in favour of keeping abstract sockets on Linux. They have a few good advantages (not having to deal with filesystem variances is one, but automatic clean-up is also quite a major one) and Linux is libqb's primary platform. The main reason that filesystem sockets were added was to overcome a limitation in the way that some container platforms behaved.

[...]

PR #339, opened 2019-03-12
ipc: use O_EXCL on SHM files

chrissie-c:

### DIFF
--- a/lib/ipc_setup.c
+++ b/lib/ipc_setup.c
@@ -43,6 +43,9 @@
 #include "util_int.h"
 #include "ipc_int.h"
 
+/* Maximum number of times we generate a random socket name before giving up */
+#define MAX_NAME_RETRY_COUNT 20
+
 struct ipc_auth_ugp {
 	uid_t uid;
 	gid_t gid;
@@ -619,6 +622,7 @@ handle_new_connection(struct qb_ipcs_service *s,
 	struct qb_ipc_connection_request *req = msg;
 	int32_t res = auth_result;
 	int32_t res2 = 0;
+	uint32_t retry_count = 0;
 	uint32_t max_buffer_size = QB_MAX(req->max_msg_size, s->max_buffer_size);
 	struct qb_ipc_connection_response response;
 
@@ -643,8 +647,6 @@ handle_new_connection(struct qb_ipcs_service *s,
 	c->auth.gid = c->egid = ugp->gid;
 	c->auth.mode = 0600;
 	c->stats.client_pid = ugp->pid;
-	snprintf(c->description, CONNECTION_DESCRIPTION,
-		 "%d-%d-%d", s->pid, ugp->pid, c->setup.u.us.sock);
 
 	if (auth_result == 0 && c->service->serv_fns.connection_accept) {
 		res = c->service->serv_fns.connection_accept(c,
@@ -657,9 +659,17 @@ handle_new_connection(struct qb_ipcs_service *s,
 	qb_util_log(LOG_DEBUG, "IPC credentials authenticated (%s)",
 		    c->description);
 
+retry_description:
+	snprintf(c->description, CONNECTION_DESCRIPTION,
+		 "%d-%d-%lu", s->pid, ugp->pid, (unsigned long)(random()%65536));
+
 	memset(&response, 0, sizeof(response));
 	if (s->funcs.connect) {
 		res = s->funcs.connect(s, c, &response);
+		if (res == -EEXIST && ++retry_count < MAX_NAME_RETRY_COUNT) {
+			qb_util_log(LOG_DEBUG, "Retrying socket name %s (count=%ld)\n", c->description, retry_count);
+			goto retry_description;
+		}
 		if (res != 0) {
 			goto send_response;
 		}
--- a/lib/ipc_setup.c
+++ b/lib/ipc_setup.c
@@ -43,9 +43,6 @@
 #include "util_int.h"
 #include "ipc_int.h"
 
-/* Maximum number of times we generate a random socket name before giving up */
-#define MAX_NAME_RETRY_COUNT 20
-
 struct ipc_auth_ugp {
 	uid_t uid;
 	gid_t gid;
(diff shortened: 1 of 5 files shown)


lib/ipc_setup.c
```diff
@@ -657,9 +659,17 @@ handle_new_connection(struct qb_ipcs_service *s,
 	qb_util_log(LOG_DEBUG, "IPC credentials authenticated (%s)",
 		    c->description);
 
+retry_description:
+	snprintf(c->description, CONNECTION_DESCRIPTION,
+		 "%d-%d-%lu", s->pid, ugp->pid, (unsigned long)(random()%65536));
```
topimiettinen: Unfortunately 16 bits of randomness is way too little, it's easy to create 64k files quickly. I'd use 64 bits as minimum. Does the protocol require a decimal number here, could it be hex or Base64 formatted instead? I'm not sure random() is a good generator of secure random numbers.

lib/ipcs.c
```diff
@@ -75,6 +77,18 @@ qb_ipcs_create(const char *name,
 	qb_list_init(&s->list);
 	qb_list_add(&s->list, &qb_ipc_services);
 
+	/* Randomise socket names */
+	fd = open("/dev/urandom", O_RDONLY);
+	if (fd == -1) {
+		seed = (time_t)time(NULL);
+	} else {
+		if (read(fd, &seed, sizeof(seed)) != 4) {
```
topimiettinen: This effectively gives 32 bits of randomness to the generator. Theoretically it's not impossible to break by brute forcing this kind of number space, though in practice it might be impossible for the attacker to generate 2^32 files in /dev/shm. I'd still try to find out a way to get at least 64 bits of randomness to the file names to be sure.

lib/ipc_setup.c
chrissie-c: Fair point

lib/ipcs.c
chrissie-c: srand() only takes an int! Should I not be using random - even with a seed from /dev/urandom?

topimiettinen: Here's a good implementation in [systemd src/basic/random-util.c](https://github.com/systemd/systemd/blob/ca66b69d5bdc9d3bf34cdc53bb600f88ef6a2c87/src/basic/random-util.c#L68), though it may need some porting to run on non-Linux. It's also a bit complex.

chrissie-c:
Thanks. It does look a bit over the top. My idea for an alternative was to read 64 bits from /dev/urandom for each connection. If that fails (which is shouldn't of course) then fall back to 2 calls to random()

jnpkrn:
Hm, and what about using POSIX 2008 endorsed `mkdtemp` and leave any
sort of uniqueness-guessing to the outter system (which can make it
truly _race-free_ when backed by the operating system), idea
I tangentially touched in the comment to the original report?

Cons are an extra system resource to consume (inode per the directory)
and care about (removal) + the necessity of wider retesting/revalidation,
but at least doesn't naively reimplement what's there to work with
already.

I.e., turn

```
/dev/shm/qb-ipcserver-request-6555-6674-9-data
```

into something like

```
/dev/shm/qb-ipcserver-request-6555-6674-9-XXXXXX/data
```

where the trailing XXXXXX corresponds to the semantics of `mkdtemp`.

Moreover, ```/dev/shm/qb-ipcserver-request-6555-6674-9-XXXXXX/```
from above would then become the main gate keeping point (plain
ACL based access control, rather than taking care of every and
each of the underlying files separately, of course, there would
be some `umask` cooperation needed for this to be the case).

lib/log_blackbox.c
```diff
@@ -188,7 +188,7 @@ qb_log_blackbox_write_to_file(const char *filename)
 	ssize_t written_size = 0;
 	struct qb_log_target *t;
 	struct _blackbox_file_header header;
-	int fd = open(filename, O_CREAT | O_RDWR, 0700);
+	int fd = open(filename, O_CREAT | O_RDWR | O_EXCL, 0700);
```
jnpkrn: `O_RDWR` shall likely be flipped to `O_WRONLY`.

`O_EXCL` is dubious, overwriting existing file is a valid use case
here and an established semantics of the function.

Perhaps devising `qb_log_blackbox_write_to_file_2` with some expressly
noted flags passable as `flags` and making `qb_log_blackbox_write_to_file`
depend on that might be a better idea.

chrissie-c:
I'm sympathetic to the idea of mkdtemp, but annoyingly it's not compatible with shm_open() as far as I can tell. FreeBSD's shm_open() call has no filesystem visibility at all. Although it does allow a '/' in filename I'm not sure if they count as a directory as such - or even what that would mean.

As we need this security patch for the 1.0.x stream are people happy with my patch as posted and modified for this branch, then we'll look at something a bit more sophisticated for the 2.0 branch?
if not let me know what needs fixing and we'll get on with it. (ish - I'm off for the rest of the week after today)

jnpkrn:
Since `shm_open` doesn't solve the pre-existing consumption problem in
any way for us, and because libqb already has one fallback path to use
intead of `/dev/shm`, I don't think there's a clear incentive to use
that.

Also, `shm_open` is not compatible with a portable discoverability
for the hypothetical passive IPC tracing/debugging tool to use
as mentioned at
https://github.com/ClusterLabs/libqb/issues/222#issuecomment-472844174
(ditto passive blackbox-backed realtime following)
-- this discoverability scheme exists _today_ (for years already), and
while it surely will get slightly ambiguous after the fix for this very
issue, it's at least still solvable on some platforms (look at
`/proc/PID/fds` etc.).

* * *

Last but not least (quite the contrary!), with your current approach,
you totally forgot that the attackers can abuse the inherent need to
create two files _in the same shared namespace_ in sequence in a way
they can observe creation of the first (`*-header`; e.g. using
`inotify`) and hurry up to win the race over the legitimate libqb
library client on creating the second one (`*-date`).  This makes
the currently proposed solution at sea.

With properly guarded intermediate directory, this wouldn't happen,
since there's a just a single entity to try to win the race over,
and that's moreover actively guarded by the operating system on
some platforms when `mkdtemp` is used.

jnpkrn:
There's WIP using `mkdtemp` and fixing thus the described shortcomings:
#343
`O_EXCL` would then be more optional, but perhaps good to use nonetheless.

Regarding the tracing tool, it'd be indeed helpful for remote debugging
of situations just before fencing event occurring (the tool run remotely
in the ssh session, it will freeze eventually, but perhaps already
capturing the most precious data) -- just recently I've observed
scenario it would certainly help with.

jnpkrn:
Sane with respect to the primary goal, yes.
Sufficient in the light of the known attack surface, no.

chrissie-c:
I'm going to shrink this patch so it only has the O_EXCL bits in. The other part can be looked at separately as it's just going to get too messy.

PR #340, opened 2019-03-13
Journal logging followup

jnpkrn:
Additional point for discussion:

- for 2.0, it would perhaps make sense to drop the priority
  stringification from the implicit message structure for
  general syslog (and blackbox) target like I do here for its
  specialization, journal target (it's all _metadata_ rather then
  explicit marking -- that's suitable for plain/raw formats only,
  IMHO)


jnpkrn:
TODO: try to get journal logging tested in Travis CI.

chrissie-c:
Cherry-picked fdb8396a383a3c0c4d07d35365cca62d1354a734

PR #342, opened 2019-03-15
Test rpm

fabbione:
allow building and shipping of libqb test suite, both from the local build and as rpm.

the patchset includes a partial example on what needs to be fixed in the test suite to run on an installed system. There are several bits that needs fixing because the test suite assume that it´s being executed from the build tree and have write access to ./

### DIFF
--- a/tests/blackbox-segfault.sh
+++ b/tests/blackbox-segfault.sh
@@ -1,6 +1,17 @@
 #!/bin/sh
 #
 # create a normal blackbox
+
+# use in-tree qb-blackbox when available
+if [ -f ../tools/qb-blackbox ]; then
+ QBBLACKBOX=../tools/qb-blackbox
+else
+ QBBLACKBOX=qb-blackbox
+fi
+
+#
+# create a normal blackbox
+#
 rm -f crash-test-dummy.fdata
 ./crash_test_dummy
 rm -f core*
@@ -9,18 +20,17 @@ rm -f core*
 
 # first test that reading the valid
 # blackbox data actually works.
-../tools/qb-blackbox crash-test-dummy.fdata
+$QBBLACKBOX crash-test-dummy.fdata
 if [ $? -ne 0 ]; then
 	exit 1
 fi
-
 
 for i in $(seq $NUM_BB_TESTS)
 do
     rm -f butchered_blackbox.fdata
     echo " ==== Corrupt blackbox test $i/$NUM_BB_TESTS ===="
     ./file_change_bytes -i crash-test-dummy.fdata -o butchered_blackbox.fdata -n 1024
-    ../tools/qb-blackbox butchered_blackbox.fdata
+    $QBBLACKBOX butchered_blackbox.fdata
     [ $? -gt 127 ] && exit 1 || true
 done
 
--- a/tests/blackbox-segfault.sh
+++ b/tests/blackbox-segfault.sh
@@ -1,26 +1,24 @@
 #!/bin/sh
 #
+# Needs PATH to be set to find accompanying test programs
+# - including qb-blackbox which for in-tree tests should be
+# - in ../tools
+#
 # create a normal blackbox
-
-# use in-tree qb-blackbox when available
-if [ -f ../tools/qb-blackbox ]; then
- QBBLACKBOX=../tools/qb-blackbox
-else
- QBBLACKBOX=qb-blackbox
-fi
+#
 
 #
 # create a normal blackbox
 #
 rm -f crash-test-dummy.fdata
-./crash_test_dummy
+crash_test_dummy
 rm -f core*
 
-. ./test.conf
+. test.conf
 
 # first test that reading the valid
 # blackbox data actually works.
-$QBBLACKBOX crash-test-dummy.fdata
+qb-blackbox crash-test-dummy.fdata
 if [ $? -ne 0 ]; then
 	exit 1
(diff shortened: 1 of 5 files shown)


jnpkrn:
> this is necessary if it´s not possible to run make check during
> make rpm

I'm lost, can you describe the use case better, please?

tests/blackbox-segfault.sh
```diff
@@ -1,6 +1,17 @@
 #!/bin/sh
 #
 # create a normal blackbox
+
+# use in-tree qb-blackbox when available
+if [ -f ../tools/qb-blackbox ]; then
+ QBBLACKBOX=../tools/qb-blackbox
+else
+ QBBLACKBOX=qb-blackbox
+fi
```
jnpkrn: Cannot `PATH` priorities handle this better?

fabbione:
On 3/15/2019 2:02 PM, Jan Pokorný wrote:
>     this is necessary if it´s not possible to run make check during
>     make rpm
> 
> I'm lost, can you describe the use case better, please?

Not all builders might be able to run the test suite (we have no control 
over the builders etc. etc. etc), hence we still want to have the 
ability to build the test suite binaries and ship them in rpm.

Fabio

fabbione:
On 3/15/2019 2:03 PM, Jan Pokorný wrote:
> *@jnpkrn* commented on this pull request.
> 
> ------------------------------------------------------------------------
> 
> In tests/blackbox-segfault.sh 
> <https://github.com/ClusterLabs/libqb/pull/342#discussion_r265970838>:
> 
>> @@ -1,6 +1,17 @@
>   #!/bin/sh
>   #
>   # create a normal blackbox
> +
> +# use in-tree qb-blackbox when available
> +if [ -f ../tools/qb-blackbox ]; then
> + QBBLACKBOX=../tools/qb-blackbox
> +else
> + QBBLACKBOX=qb-blackbox
> +fi
> 
> Cannot |PATH| priorities handle this better?

Quoting the very first part of the PR:

"the patchset includes a partial example on what needs to be fixed in 
the test suite to run on an installed system. There are several bits 
that needs fixing because the test suite assume that it´s being executed 
from the build tree and have write access to ./"

This is just a hint on what needs fixing. the "how" to fix the whole 
test suite is out of scope from this PR.

Fabio

chrissie-c:
Also, a key feature is being able to run the tests on the installed version of libqb, and not on a build-tree version.

jnpkrn:
I see the big picture, totally makes sense and thanks for that effort.
Less so step-wise if I wasn't a devoted person and trying to make
sense from the commit messages alone.

Please, mark the PR as "not to be merged yet" (there's an option
to be checked by the originator, IIRC) if the current version is
more a brainstorm session with production quality to arrive later
incl. rather rushed intermittent commit messages, so it won't 
trigger this sort of feedback (assuming this would be the final
version) -- sorry if that's the case.

This really explains the matter better:

```
Currently, tests binaries are only compiled upon "make check".
They can be, though, required in a compiled form indepedently of
running them.  Solution is to make them compiled unconditionally.
```

Which I am not fully content with, actually, I think it'd be better
if they continued not compiled by default, only when configured with
`--with-tests` (which would error out if `libcheck` is not around).
This would be fully isolated from the `make check` use case, I don't
think it would conflict with anything if the same set of executables
was listed under two variables (`test_PROGRAMS` sourcing from
`check_PROGRAMS`, possibly), one automake-conditionalized respective
to whether `--with-tests` was passed to `configure` or not.

Would this arrangement be sensible for the projected use in the
builders and wherever else?

fabbione:
On 3/15/2019 5:52 PM, Jan Pokorný wrote:
> I see the big picture, totally makes sense and thanks for that effort.
> Less so step-wise if I wasn't a devoted person and trying to make
> sense from the commit messages alone.
> 
> Please, mark the PR as "not to be merged yet" (there's an option
> to be checked by the originator, IIRC) if the current version is
> more a brainstorm session with production quality to arrive later
> incl. rather rushed intermittent commit messages, so it won't
> trigger this sort of feedback (assuming this would be the final
> version) -- sorry if that's the case.

Only the last commit is partial as already mentioned. The other commits 
are good to go in as-is (code wise, I can agree the commit message could 
be more verbose).

> 
> This really explains the matter better:
> 
> |Currently, tests binaries are only compiled upon "make check". They can 
> be, though, required in a compiled form indepedently of running them. 
> Solution is to make them compiled unconditionally. |
> 
> Which I am not fully content with, actually, I think it'd be better
> if they continued not compiled by default, only when configured with
> |--with-tests| (which would error out if |libcheck| is not around).

Makes absolutely no difference to build it by default or with yet 
another option. this is splitting the hair for what use case exactly? 
what problem does it solve or not solve to build them only --with-tests, 
except having to add more build options?

> This would be fully isolated from the |make check| use case, I don't
> think it would conflict with anything if the same set of executables
> was listed under two variables (|test_PROGRAMS| sourcing from
> |check_PROGRAMS|, possibly), one automake-conditionalized respective
> to whether |--with-tests| was passed to |configure| or not.
> 
> Would this arrangement be sensible for the projected use in the
> builders and wherever else?

Feel free to make your version of the PR, but it has to be done ASAP 
because we need propagation of the rpm around.

Fabio

jnpkrn:
Upstream development is not about whether it works for a narrow group
of people.  Please, respect that.

PR #343, opened 2019-03-22
WIP: use mkdtemp for IPC paths

jnpkrn:
work in progress:

* [x] core implementation
* [ ] code cleanup, doc, valgrind sanitization etc.
* [ ] `chmod` + `chown` recursion, also reconsider the permissions
       scheme (single perms guard at the dir + lax approach for items,
       vs. everything strict) and sanity checking
       (`data` contained within the same parent dir `header`
       enforcement in client)
* [ ] `rmdir` at more places as a proper cleanup 
* [ ] using `gnulib` surrogate for `mkdtemp` where not available natively

### DIFF
--- a/lib/ipc_shm.c
+++ b/lib/ipc_shm.c
@@ -309,32 +309,49 @@ qb_ipcs_shm_connect(struct qb_ipcs_service *s,
 		    struct qb_ipc_connection_response *r)
 {
 	int32_t res;
+	char request[PATH_MAX], response[PATH_MAX], event[PATH_MAX];
+	char *where_to_affix;
 
 	qb_util_log(LOG_DEBUG, "connecting to client [%d]", c->pid);
 
-	snprintf(r->request, NAME_MAX, "%s-request-%s",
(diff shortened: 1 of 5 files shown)


radosroka:
Hi @jnpkrn , is this also going to fix #369 ?

jnpkrn:
> is this also going to fix #369?

For explanation, this was to demonstrate (mainly to @chrissie-c who
originally set off in the ill-devised direction):

https://github.com/ClusterLabs/libqb/pull/339#issuecomment-475754091

@radosroka:

You apparently struggle with phenomenon of directory ownership change
(that you seem would expect not to be mangled in contrast how it would
have been by standard process-to-objects permissions transfer), but
rest assure this PR is definitely not going to help (if anything, it
does _not_ handle that problem at all -- while it was apparent it needs
some sort of assessment should it be eventually picked -- see the TODO
notes:

> - chmod + chown, also reconsider the permissions scheme

That being said, closing this PR so as not to cast any confusions.

PR #345, opened 2019-04-04
ipc: Use mkdtemp for more secure IPC files

chrissie-c:
This is not meant to be the last word on the subject, it's mainly a
simple way of making the current libqb more secure. Importantly, it's backwards
compatible with an old server.

It calls rmdir on the directory created by mkdtemp *way* too often, but
it seems to be the only way to be sure that things get cleaned up on
the various types of server/client exit. I'm sure we can come up with
something tidier for master but I hope this, or something similar, will
be OK for 1.0.x.

The O_EXCL patch is also still recommended, of course.

### DIFF
--- a/lib/ipc_setup.c
+++ b/lib/ipc_setup.c
@@ -643,8 +643,23 @@ handle_new_connection(struct qb_ipcs_service *s,
 	c->auth.gid = c->egid = ugp->gid;
 	c->auth.mode = 0600;
 	c->stats.client_pid = ugp->pid;
+
+#if defined(QB_LINUX) || defined(QB_CYGWIN)
+	snprintf(c->description, CONNECTION_DESCRIPTION,
+		 "/dev/shm/qb-%d-%d-%d-XXXXXX", s->pid, ugp->pid, c->setup.u.us.sock);
+	if (mkdtemp(c->description) == NULL) {
+		res = errno;
+		goto send_response;
+	}
+
+	/* We can't pass just a directory spec to the clients */
+	strncat(c->description,"/qb", CONNECTION_DESCRIPTION);
+#else
 	snprintf(c->description, CONNECTION_DESCRIPTION,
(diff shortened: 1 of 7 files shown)


fabbione:
patch looks sane to me.

chrissie-c:
retest this please

chrissie-c:
Thanks Honza. I've added a chown() call which fixes it.

chrissie-c:
Committed as a single commit 6a4067c1d1764d93d255eccecfd8bf9f43cb0b4d

lib/ipc_setup.c
```diff
@@ -865,3 +880,22 @@ qb_ipcs_us_connection_acceptor(int fd, int revent, void *data)
 	qb_ipcs_uc_recv_and_auth(new_fd, s);
 	return 0;
 }
+
+void remove_tempdir(const char *name, size_t namelen)
+{
+#if defined(QB_LINUX) || defined(QB_CYGWIN)
+	char dirname[PATH_MAX];
+	char *slash;
+	memcpy(dirname, name, namelen);
```
wferi: Wouldn't `QB_MIN(nameline, sizeof dirname - 1)` be safer instead of `namelen` here (to reserve the final byte for a manual NUL)? Or a straight `strlcpy()`? Having `namelen` passed in raises doubts whether `name` is expected to be zero-terminated or not. The current uses are safe, though, except for the `strncat()` in `lib/ipc_setup.c`, which can overwrite the `description` field with a single byte (the closing NUL) and consequently leave the field unterminated.

wferi: Actually, that `strncat()` can overwrite 3 (the length of its `src` argument, because it's smaller than `CONNECTION_DESCRIPTION`.

chrissie-c:
ooh good catch. We should probably include this in the permissions fix as it's all part of the same regression

wferi:
Opened #349 for this. Somewhat long-winded...

Issue #338, opened 2019-03-01
Insecure Temporary Files


topimiettinen:
Libqb creates files in world-writable directories (`/dev/shm`, `/tmp`) with rather predictable file names (e.g. `/dev/shm/qb-usbguard-request-7096-835-12-data` in case of USBGuard). Also `O_EXCL` flag is not used when opening the files. This could be exploited by a local attacker to overwrite privileged system files (if not restricted by sandboxing, MAC or symlinking policies).

At least `O_EXCL` flag should be used. I'd also use more complex logic where files are created with unpredictable names (also using `O_TMPFILE`) and then possibly renamed to match file naming convention (if the protocol does not allow completely random file names). I would not use files for IPC.

topimiettinen:
I've opened issue USBGuard/usbguard#277 for USBGuard.

jnpkrn:
Thank you very much, Topi, for taking time to report this, initiating
hence a wider investigation.

(Cat is out of the bag by now, and GitHub cannot make the issues private
anyway, AFAIK.)

TBH, this got me really worried that I should have pursued the initial
suspicion at the first opportunity (not very trivial counterproof in
the legacy -- as in the original authors no longer around -- software
where the consequences of a change may not be straightforward was one
of the inhibitors), and sadly, this indeed has bad implications
(for anyone, whenever you suspect something like that, please follow
https://wiki.clusterlabs.org/wiki/Security#Reporting that's the best
genetic way to report things like that, since there's no particular
contact is dedicated for dealing with handling such reports within
the scope of ClusterLabs projects incl. libqb).

Regarding the possible quick workaround at the client side, commented
(with negative statement) at mentioned USBGuard/usbguard#277.

What works with Linux kernel 4.19+ (or perhaps through distro backports)
instantly, though is:

```
cat /proc/sys/fs/protected_regular && echo 1 > /proc/sys/fs/protected_regular
```

This is deemed a good enough _immediate_ workaround where supported
(for instance, systemd installation may enable this by default,
as is the case with Fedora).

Patch following the suggestion above to come once several things gets
considered (alternative is to enclose the files into a dedicated,
assurably unique and well guarded subdir of `/dev/shm`).

topimiettinen:
In addition to `fs.protected_regular`, I'd also use related sysctls
```
fs.protected_fifos = 1
fs.protected_hardlinks = 1
fs.protected_symlinks = 1
```

[...]

PR #347, opened 2019-04-17
Fix directory modes under /dev/shm and tolerate some lack of permissions

wferi:
This was a regression introduced by #338.


chrissie-c:
Thank you so much for this. I'll do some more testing and get a 1.0.5 out after Easter.

wferi:
This was included in #349.

PR #348, opened 2019-04-18
ipc: Allow chown to fail, we might not be running as root,

chrissie-c:
also set perms of the directory created by mkdtemp


chrissie-c:
Closed in favour of #347

PR #349, opened 2019-04-18
Check for directory name truncation and provide safe removal

wferi:
The first part of this is actually PR #347, the last two commits are new.


chrissie-c:
Thanks !

PR #350, opened 2019-04-28
Maximum length of qbarray (documentation)

wferi:
I find the parentheses too elaborate for a typo, but couldn't think of anything better.
The second commit is mostly a test of my understanding.


chrissie-c:
Thanks. There's a lot of rather bizarre English in the comments and man pages that need looking at over time, we'll get through it !

Issue #325, opened 2018-09-26
[new API] two-phased qb_ipcs_create required


jnpkrn:
Signatures
```
/* Creates a listening fd for a UNIX socket, akin to qb_ipcs_create */
qb_ipcs_listener_t* qb_ipcs_listener(const char *name);

/* Reuses existing UNIX sockets, presumably created with qb_ipcs_listener
   (e.g. the name gets validated), and extracts "name" from it using
   getsockname(2);  qb_ipcs_listener + qb_ipcs_create_from_listener
   equals qb_ipcs_create */
qb_ipcs_service_t* qb_ipcs_create_from_listener(qb_ipcs_listener *l,
                                                int32_t service_id,
                                                enum qb_ipc_type type,
                                                struct qb_ipcs_service_handlers *handlers);
```

jnpkrn:
Note that this would also (finally, one wants to say) offer client
programs to _unset_ `close-on-exec` flag for cases it would counter
productive (what if pre-fork parent establishes the listening socket
only to be picked, with `qb_ipcs_listener_from_fd`, by an after-fork
child with a brand new process image that was just exec'd.

I'm currently split whether it should be a `QB` branded flag that can
be passed to `qb_ipcs_listener` for convenience or whether authors
of client programs should be kept on their own in this.

Nonetheless, it must be documented that normally, that socket is
preconfigured with O_NONBLOCK and FD_CLOEXEC flags (respective to
particular `fcntl` command).  And that only FD_CLOEXEC can reasonably
be mangled with, everything else asks for conflicting with intended
state of affairs on libqb side.

jfriesse:
API looks sane, but what is the use case?

Also I was unable to understand following part:
> btw. this looks exactly like a sequence every forking daemon
> utilizing libqb's IPC subsystem should exercise unless parent-child
> synchronization is in place -- it's highly undesirable for daemon
> launcher process to indicate successful startup when the forked off
> executive part fail on already occupied IPC channel the moment later!).

If it is about daemonization process, then why should daemon do the exec after fork?

I can imagine situation with child processes, but sharing fd is just asking for a big troubles.

jnpkrn:
On 27/09/18 00:18 -0700, Jan Friesse wrote:
> If it is about daemonization process, then why should daemon do the
> exec after fork?

In case it's a hierarchical daemon of daemons (e.g. pacemaker).

> I can imagine situation with child processes, but sharing fd is just
> asking for a big troubles.

[...]

PR #352, opened 2019-05-16
IPC: server: avoid temporary channel priority loss, up to deadlock-worth

jnpkrn:
It turns out that while 7f56f58 allowed for less blocking (thus
throughput increasing) initial handling of connections from clients
within the abstract (out-of-libqb managed) event loop, it unfortunately
subscribes itself back to such polling mechanism for UNIX-socket-check
with a default priority, which can be lower than desired (via explicit
qb_ipcs_request_rate_limit() configuration) for particular channel
(amongst attention-competing siblings in the pool, the term here
refers to associated communication, that is, both server and
on-server abstraction for particular clients).

On top of that, it violates the natural assumption that once (single
threaded, which is imposed by libqb, at least between initial accept()
and after-said-UNIX-socket-check) server accepts the connection, it
shall rather take care of serving it (at least within stated initial
scope of client connection life cycle) rather than be rushing to accept
 new ones -- which is exactly what used to happen previously once the
library user set the effectively priority in the abstract poll
above the default one.

It's conceivable, just as with the former case of attention-competing
siblings with higher priority whereby they could _infinitely_ live on
at the expense of starving the client in the initial handling phase
(authentication) despite the library user's as-high-as-siblings
intention (for using the default priority for that unconditionally
instead, which we address here), the dead lock is imminent also in
this latter accept-to-client-authentication-handling case as well
if there's an _unlimited_ fast-paced arrival queue (well, limited
by with number of allowable open descriptors within the system,
but for the Linux built-in maximum of 1M, there may be no practical
difference, at least for time-sensitive applications).

The only hope then is that such dead-locks are rather theoretical,
since a "spontaneous" constant stream of either communication on
unrelated, higher-prio sibling channels, or of new connection arrivals
can as well testify the poor design of the libqb's IPC application.
That being said, unconditional default priority in the isolated
context of initial server-side client authentication is clearly
a bug, but such application shall apply appropriate rate-limiting
measures (exactly on priority basis) to handle unexpected flux
nonetheless.

### DIFF
--- a/include/qb/qbloop.h
+++ b/include/qb/qbloop.h
@@ -36,6 +36,26 @@ extern "C" {
  *
  * Main loop manages timers, jobs and polling sockets.
  *
+ * Only a weaker sense of priorities is implemented, alluding to distinct
+ * set of pros and cons compared to the stronger, strict approach to them
+ * as widely applied in this problem space (since the latter gives the
+ * application more control as the effect of the former can still be
+ * achieved with some reductions, whereas it is not straightforward the
+ * other way around; cf. static priority task scheduling vs. relative
+ * fine-tuning within a single priority domain with nice(2)):
+ *
+ * + implicit mitigation for deadlock-prone priority arrangements
+ *
+ * - less predictable (proportional probability based, we can talk
+ *   about an advisory effect of the priorities) responses to the arrival
+ *   of the high-ranked events (i.e. in the process of the picking the next
+ *   event to handle from the priority queue when at least two different
+ *   priorities are eligible at the moment)
+ *
+ * One practical application for this module of libqb is in combination with
+ * IPC servers based on qbipcs.h published one (the #qb_ipcs_poll_handlers
+ * structure maps fittingly to the control functions published here).
+ *
  * @example tcpserver.c
  */
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ addons:
                 packages:
                         - check
                         - splint
+                        # for ipc.test:test_ipc_dispatch_*_deadlock_provoke
+                        - libglib2.0-dev  # natively present, but doesn't hurt
                         # for "make rpm"
                         - doxygen
                         - rpm
--- a/include/qb/qbipcs.h
+++ b/include/qb/qbipcs.h
@@ -43,6 +43,12 @@ extern "C" {
  * @example ipcserver.c
  */
 
+/**
+ * Rates to be passed to #qb_ipcs_request_rate_limit.  The exact interpretation
+ * depends on how the event loop implementation understands the concept of
+ * priorities, see the discussion at #qb_ipcs_poll_handlers structure -- an
+ * integration point between IPC server instance and the underlying event loop.
+ */
 enum qb_ipcs_rate_limit {
 	QB_IPCS_RATE_FAST,
 	QB_IPCS_RATE_NORMAL,
@@ -104,6 +110,22 @@ typedef int32_t (*qb_ipcs_job_add_fn)(enum qb_loop_priority p,
 				      void *data,
 				      qb_loop_job_dispatch_fn dispatch_fn);
 
+/**
+ * A set of callbacks that need to be provided (only #job_add can be #NULL)
+ * whenever the IPC server is to be run (by the means of #qb_ipcs_run).
+ * It is possible to use accordingly named functions defined in qbloop.h module
+ * or integrate with other existing (like GLib's event loop) or entirely new
+ * code -- see the subtle distinction amongst the possible event loops pointed
+ * out in the introductory comment at qbloop.h.
+ *
+ * At that occasion, please note the correlation of #QB_IPCS_RATE_FAST etc.
+ * symbolic names with said advisory effect of the priorities in the native
+ * implementation.  This correspondence will not be this intuitively seemless
+ * if some other event loop implementation is hooked in given that it abids
+ * them strictly as mentioned (e.g. GLib's event loop over poll'able sources).
+ * Differences between the two paradigms should also be accounted for when
+ * the requirement to swap the event loop implementations arises.
+ */
 struct qb_ipcs_poll_handlers {
 	qb_ipcs_job_add_fn job_add;
 	qb_ipcs_dispatch_add_fn dispatch_add;
--- a/lib/ipc_setup.c
+++ b/lib/ipc_setup.c
@@ -825,10 +825,10 @@ qb_ipcs_uc_recv_and_auth(int32_t sock, struct qb_ipcs_service *s)
 	setsockopt(sock, SOL_SOCKET, SO_PASSCRED, &on, sizeof(on));
 #endif
 
-	res = s->poll_fns.dispatch_add(QB_LOOP_MED,
-					data->sock,
-					POLLIN | POLLPRI | POLLNVAL,
-					data, process_auth);
+	res = s->poll_fns.dispatch_add(s->poll_priority,
+	                               data->sock,
+	                               POLLIN | POLLPRI | POLLNVAL,
+	                               data, process_auth);
 	if (res < 0) {
 		qb_util_log(LOG_DEBUG, "Failed to process AUTH for fd (%d)", data->sock);
 		close(sock);
--- a/lib/ipc_setup.c
+++ b/lib/ipc_setup.c
@@ -830,7 +830,8 @@ qb_ipcs_uc_recv_and_auth(int32_t sock, struct qb_ipcs_service *s)
 	                               POLLIN | POLLPRI | POLLNVAL,
 	                               data, process_auth);
 	if (res < 0) {
-		qb_util_log(LOG_DEBUG, "Failed to process AUTH for fd (%d)", data->sock);
+		qb_util_log(LOG_DEBUG, "Failed to arrange for AUTH for fd (%d)",
+		            data->sock);
 		close(sock);
 		destroy_ipc_auth_data(data);
 	}
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -167,6 +167,11 @@ ipc_test_LDADD = $(top_builddir)/lib/libqb.la @CHECK_LIBS@
 if HAVE_FAILURE_INJECTION
 ipc_test_LDADD += _failure_injection.la
 
+if HAVE_GLIB
+ipc_test_CFLAGS += $(GLIB_CFLAGS)
+ipc_test_LDADD += $(GLIB_LIBS)
+endif
+
 check_LTLIBRARIES += _failure_injection.la
 _failure_injection_la_SOURCES = _failure_injection.c _failure_injection.h
 _failure_injection_la_LDFLAGS = -module
--- a/tests/check_ipc.c
+++ b/tests/check_ipc.c
@@ -468,8 +468,10 @@ NEW_PROCESS_RUNNER(run_ipc_server, ready_signaller, signaller_data)
 }
 
 static pid_t
-run_function_in_new_process(new_process_runner_fn new_process_runner)
+run_function_in_new_process(const char *role,
+                            new_process_runner_fn new_process_runner)
 {
+	char formatbuf[1024];
 	pid_t parent_target, pid1, pid2;
 	struct sigaction orig_sa, purpose_sa;
 	sigset_t orig_mask, purpose_mask, purpose_clear_mask;
@@ -496,6 +498,15 @@ run_function_in_new_process(new_process_runner_fn new_process_runner)
 		}
 		if (pid2 == 0) {
 			sigprocmask(SIG_SETMASK, &orig_mask, NULL);
+
+			if (role == NULL) {
+				qb_log_format_set(QB_LOG_STDERR, "lib/%f|%l[%P] %b");
+			} else {
+				snprintf(formatbuf, sizeof(formatbuf),
+				         "lib/%%f|%%l|%s[%%P] %%b", role);
+				qb_log_format_set(QB_LOG_STDERR, formatbuf);
+			}
+
 			new_process_runner(usr1_signaller, &parent_target);
 			exit(0);
 		} else {
@@ -664,7 +675,7 @@ test_ipc_txrx_timeout(void)
 	pid_t pid;
 	uint32_t max_size = MAX_MSG_SIZE;
 
-	pid = run_function_in_new_process(run_ipc_server);
+	pid = run_function_in_new_process("server", run_ipc_server);
 	fail_if(pid == -1);
 
 	do {
@@ -712,7 +723,7 @@ test_ipc_txrx(void)
 	pid_t pid;
 	uint32_t max_size = MAX_MSG_SIZE;
 
-	pid = run_function_in_new_process(run_ipc_server);
+	pid = run_function_in_new_process("server", run_ipc_server);
 	fail_if(pid == -1);
 
 	do {
@@ -762,7 +773,7 @@ test_ipc_exit(void)
 	pid_t pid;
 	uint32_t max_size = MAX_MSG_SIZE;
 
-	pid = run_function_in_new_process(run_ipc_server);
+	pid = run_function_in_new_process("server", run_ipc_server);
 	fail_if(pid == -1);
 
 	do {
@@ -924,7 +935,7 @@ test_ipc_dispatch(void)
 	int32_t size;
 	uint32_t max_size = MAX_MSG_SIZE;
 
-	pid = run_function_in_new_process(run_ipc_server);
+	pid = run_function_in_new_process("server", run_ipc_server);
 	fail_if(pid == -1);
 
 	do {
@@ -1051,7 +1062,7 @@ test_ipc_stress_connections(void)
 			  QB_LOG_FILTER_FILE, "*", LOG_INFO);
 	qb_log_ctl(QB_LOG_STDERR, QB_LOG_CONF_ENABLED, QB_TRUE);
 
-	pid = run_function_in_new_process(run_ipc_server);
+	pid = run_function_in_new_process("server", run_ipc_server);
 	fail_if(pid == -1);
 
 	for (connections = 1; connections < 70000; connections++) {
@@ -1098,7 +1109,7 @@ test_ipc_bulk_events(void)
 	int32_t fd;
 	uint32_t max_size = MAX_MSG_SIZE;
 
-	pid = run_function_in_new_process(run_ipc_server);
+	pid = run_function_in_new_process("server", run_ipc_server);
 	fail_if(pid == -1);
 
 	do {
@@ -1162,7 +1173,7 @@ test_ipc_stress_test(void)
 	int32_t real_buf_size;
 
 	enforce_server_buffer = 1;
-	pid = run_function_in_new_process(run_ipc_server);
+	pid = run_function_in_new_process("server", run_ipc_server);
 	enforce_server_buffer = 0;
 	fail_if(pid == -1);
 
@@ -1264,7 +1275,7 @@ test_ipc_event_on_created(void)
 
 	num_bulk_events = 1;
 
-	pid = run_function_in_new_process(run_ipc_server);
+	pid = run_function_in_new_process("server", run_ipc_server);
 	fail_if(pid == -1);
 
 	do {
@@ -1318,7 +1329,7 @@ test_ipc_disconnect_after_created(void)
 	int32_t res;
 	uint32_t max_size = MAX_MSG_SIZE;
 
-	pid = run_function_in_new_process(run_ipc_server);
+	pid = run_function_in_new_process("server", run_ipc_server);
 	fail_if(pid == -1);
 
 	do {
@@ -1375,7 +1386,7 @@ test_ipc_server_fail(void)
 	pid_t pid;
 	uint32_t max_size = MAX_MSG_SIZE;
 
-	pid = run_function_in_new_process(run_ipc_server);
+	pid = run_function_in_new_process("server", run_ipc_server);
 	fail_if(pid == -1);
 
 	do {
@@ -1509,7 +1520,7 @@ test_ipc_service_ref_count(void)
 
 	reference_count_test = QB_TRUE;
 
-	pid = run_function_in_new_process(run_ipc_server);
+	pid = run_function_in_new_process("server", run_ipc_server);
 	fail_if(pid == -1);
 
 	do {
--- a/tests/check_ipc.c
+++ b/tests/check_ipc.c
@@ -926,22 +926,19 @@ struct my_res {
 	char message[1024 * 1024];
 };
 
-static void
-test_ipc_dispatch(void)
+static inline
+NEW_PROCESS_RUNNER(client_dispatch, ready_signaller, signaller_data)
 {
-	int32_t j;
-	int32_t c = 0;
-	pid_t pid;
-	int32_t size;
 	uint32_t max_size = MAX_MSG_SIZE;
-
-	pid = run_function_in_new_process("server", run_ipc_server);
-	fail_if(pid == -1);
+	int32_t size;
+	int32_t c = 0;
+	int32_t j;
+	pid_t server_pid = *((pid_t *) signaller_data);
 
 	do {
 		conn = qb_ipcc_connect(ipc_name, max_size);
 		if (conn == NULL) {
-			j = waitpid(pid, NULL, WNOHANG);
+			j = waitpid(server_pid, NULL, WNOHANG);
 			ck_assert_int_eq(j, 0);
 			poll(NULL, 0, 400);
 			c++;
@@ -949,16 +946,31 @@ test_ipc_dispatch(void)
 	} while (conn == NULL && c < 5);
 	fail_if(conn == NULL);
 
+	if (ready_signaller != NULL) {
+		ready_signaller(signaller_data);
+	}
+
 	size = QB_MIN(sizeof(struct qb_ipc_request_header), 64);
 	for (j = 1; j < 19; j++) {
 		size *= 2;
 		if (size >= max_size)
 			break;
 		if (send_and_check(IPC_MSG_REQ_DISPATCH, size,
-				   recv_timeout, QB_TRUE) < 0) {
+		                   recv_timeout, QB_TRUE) < 0) {
 			break;
 		}
 	}
+}
+
+static void
+test_ipc_dispatch(void)
+{
+	pid_t pid;
+
+	pid = run_function_in_new_process(NULL, run_ipc_server);
+	fail_if(pid == -1);
+
+	client_dispatch(NULL, (void *) &pid);
 
 	request_server_exit();
 	qb_ipcc_disconnect(conn);
--- a/tests/check_ipc.c
+++ b/tests/check_ipc.c
@@ -399,8 +399,30 @@ s1_connection_created(qb_ipcs_connection_t *c)
 
 }
 
-static void
-run_ipc_server(void)
+static volatile sig_atomic_t usr1_bit;
+
+static void usr1_bit_setter(int signal) {
+    if (signal == SIGUSR1) {
+        usr1_bit = 1;
+    }
+}
+
+#define READY_SIGNALLER(name, data_arg)  void (name)(void *data_arg)
+typedef READY_SIGNALLER(ready_signaller_fn, );
+
+static
+READY_SIGNALLER(usr1_signaller, parent_target)
+{
+	kill(*((pid_t *) parent_target), SIGUSR1);
+}
+
+#define NEW_PROCESS_RUNNER(name, ready_signaller_arg, signaller_data_arg) \
+	void (name)(ready_signaller_fn ready_signaller_arg, \
+	      void *signaller_data_arg)
+typedef NEW_PROCESS_RUNNER(new_process_runner_fn, , );
+
+static
+NEW_PROCESS_RUNNER(run_ipc_server, ready_signaller, signaller_data)
 {
 	int32_t res;
 	qb_loop_signal_handle handle;
@@ -423,7 +445,7 @@ run_ipc_server(void)
 
 	my_loop = qb_loop_create();
 	qb_loop_signal_add(my_loop, QB_LOOP_HIGH, SIGTERM,
-			   NULL, exit_handler, &handle);
+	                   NULL, exit_handler, &handle);
 
 
 	s1 = qb_ipcs_create(ipc_name, 4, ipc_type, &sh);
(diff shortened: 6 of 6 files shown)


jnpkrn:
Not sure why kronosnet CI is not passing on FreeBSD 12 (`ipc.test`).
It worked for me without a glitch with `12.0-RELEASE r341666` in VM.

jnpkrn:
It turns out that the bug is effectively unspottable with libqb's na(t)ive
event loop implementation (`qbloop.h`).  Naive, for it doesn't even
implement priority levels properly[*], making broken SW behave as if
everything was alright :-/

[*] actually it's too harsh, since it's unstated whether this implementation
    is this forgiving on purpose (although for soft real-time purposes, it'd
    be stretch to call this a desired property), and it does some sort of
    (again naive) prioritity balancing

Do not merge yet, the new unit test currently errs on the safe side, marking
a failure even after the fix -- need to find some reliable/reproducible
balance.

jnpkrn:
Honestly, this was the most painful unit test ever for me.

@chrissie-c if you are OK with this, I'd prefer pushing it myself.

include/qb/qbloop.h
```diff
@@ -36,6 +36,16 @@ extern "C" {
  *
  * Main loop manages timers, jobs and polling sockets.
  *
+ * To be noted that local implementation of the event loop does care about
+ * priorities, these are rather advisory, however.  For the true real-world
+ * systems with requirements of deterministic responses, you may be better
+ * served with these priorities strictly separated and abided, which is the
+ * case, e.g., with GLib.  On the other hand, for early stages and timing
```
jfriesse: I don't really have too high opinion about libqb loop, but it may be helpful to be a little more specific. What do you mean by "... are rather advisory"? Why it doesn't fulfill "deterministic" guarantees (and maybe a small note why Glib does so?)? Also are you sure that your advice is really true for all (or at least most of the) "real-world" systems?

jnpkrn:
On 28/05/19 02:03 -0700, Jan Friesse wrote:
> jfriesse commented on this pull request.
> I don't really have too high opinion about libqb loop, but it may be
> helpful to be a little more specific. What do you mean by "... are
> rather advisory"?

It's basically a hint rather than strict priority ordering abided
when handling order of the concurrent events is to be decided.
AFAIK, for the three priorities available, the ratio of serving
them is 3:2:1 -- it's just a probabilistic method of dealing with
priorities, which is rather atypical, and makes (high-rate) high-prio
batch event stream possibly blocked due to slow to handle (problem
of the application, but nevertheless...) low prio events.

> Why it doesn't fulfill "deterministic" guarantees (and maybe
> a small note why Glib does so?)?

Deterministic ("absolutistic") as opposed to probabilistic
(inherently nondeterministic), see above.

If you want to take me too literally, of course, when there's
the only event source (at possessions of both differently prioritized
channels) for the event handler, it's still deterministic, since this
source has a full control of the requests (until they are running
fully asynchronously), but it's gone when there are independent
sources.

> Also are you sure that your advice is really true for all (or at
> least most of the) "real-world" systems?

I am actually not aware of systems where probabilistic approach to
priorities would get used (apart from libqb).  Hand-written loop
management is also typically based on implicit, strict priorities
(statically hard-coded order of handling the particular events).

The experience (see the unit test) tells that GLib behaves in what
I refer to as deterministic manner.  That's somewhat more intuitive,
yet more demanding when it comes to the correct usage (requiring
starvation/deadlock-avoiding design considerations).

It's basically "process niceness vs. real-time scheduling" analogy.
Libqb favours the former (without declaring that clearly, until
the commit in question), GLib et al. the latter.

Any suggestions how to word this better while still guiding the libqb
users towards correct expectations welcome.

-- 
Jan (Poki)

jnpkrn:
Reworded and extended the commentary in `qbloop.h`, fixing a typo along.
Honestly, I have no idea to what extent the non-pollable sources in
GLib's mainloop are subject to priority-based serialization, so narrowed
the stated example down just to that.  Also added (currently extraneous)
prerequisite on GLib, so that those new tests won't get accidentally
avoided.

include/qb/qbloop.h
chrissie-c: The wording there is hard to understand for me, it's not in idiomatic modern English. I'll work out a better phrasing and let you know what I come up with.

jfriesse:
> On 28/05/19 02:03 -0700, Jan Friesse wrote: jfriesse commented on this pull request. I don't really have too high opinion about libqb loop, but it may be helpful to be a little more specific. What do you mean by "... are rather advisory"?
> It's basically a hint rather than strict priority ordering abided when handling order of the concurrent events is to be decided. AFAIK, for the three priorities available, the ratio of serving them is 3:2:1 -- it's just a probabilistic method of dealing with priorities, which is rather atypical, and makes (high-rate) 

Ok, but is it necessarily bad?

> high-prio batch event stream possibly blocked due to slow to handle (problem of the application, but nevertheless...) low prio events.

Yep. So why not to describe it and let user to choose? Maybe this behavior is better fit?

> Why it doesn't fulfill "deterministic" guarantees (and maybe a small note why Glib does so?)?
> Deterministic ("absolutistic") as opposed to probabilistic (inherently nondeterministic), see above. If you want to take me too literally, of course, when there's the only event source (at possessions of both differently prioritized channels) for the event handler, it's still deterministic, since this source has a full control of the requests (until they are running fully asynchronously), but it's gone when there are independent sources.

Current loop is as deterministic as glib implementation. It's just different behavior.

> Also are you sure that your advice is really true for all (or at least most of the) "real-world" systems?
> I am actually not aware of systems where probabilistic approach to priorities would get used (apart 

So no hard data

> from libqb). Hand-written loop management is also typically based on implicit, strict priorities (statically 

Do you have any hard data?

> hard-coded order of handling the particular events). The experience (see the unit test) tells that GLib behaves in what I refer to as deterministic manner. That's somewhat more intuitive, yet more demanding 

It's easier to describe glib behavior, but I wouldn't call it more deterministic

> when it comes to the correct usage (requiring starvation/deadlock-avoiding design considerations). It's basically "process niceness vs. real-time scheduling" analogy. Libqb favours the former (without 

Kind of right analogy and maybe nice to have it in description,.

declaring that clearly, until the commit in question), GLib et al. the latter. Any suggestions how to word this better while still guiding the libqb users towards correct expectations welcome.

Because chrissie is on it I will not add my description but idea is to really describe how current implementation works and what may be drawbacks. That's it. No need to describe feeling and/or give advice without any real evidence.

> […](#)
> -- Jan (Poki)

chrissie-c:
My idea for working is much simpler and to the point.

"The priorities passed to the libqb loop functions are advisory and provide no guarantees as to the order the callbacks will be invoked."

I don't think we need to say any more than that. Offering comparisons with other libraries just runs the risk of things getting out of date.

jfriesse:
@chrissie-c Yep, this short description serve probably as well as more in-dept" description of current implementation, so ACK.

jnpkrn:
Such a cross-fire about the comment.

On 29/05/19 01:09 -0700, Jan Friesse wrote:
>> On 28/05/19 02:03 -0700, Jan Friesse wrote:
>>> I don't really have too high opinion about libqb loop, but it may
>>> be helpful to be a little more specific. What do you mean by "...
>>> are rather advisory"?
>>
>> It's basically a hint rather than strict priority ordering abided
>> when handling order of the concurrent events is to be decided.
>> AFAIK, for the three priorities available, the ratio of serving
>> them is 3:2:1 -- it's just a probabilistic method of dealing with
>> priorities, which is rather atypical, and makes (high-rate) 
> 
> Ok, but is it necessarily bad?

Definitely not.  It's explained that easy way out of starvation
problems.  It's surprising, though, and I think this fact deserves
to be reflected.

>> high-prio batch event stream possibly blocked due to slow to handle
>> (problem of the application, but nevertheless...) low prio events.
> 
> Yep. So why not to describe it and let user to choose? Maybe this
> behavior is better fit?

Users have the liberty of choice, there's no doubt about that, and
when they decide to stick with libqb's native implementation, they
should be perfectly aware of the trade-off they make when doing so.
If they are not OK with that, they will better be served elsewhere.

>>> Why it doesn't fulfill "deterministic" guarantees (and maybe a
>>> small note why Glib does so?)?
>> 
>> Deterministic ("absolutistic") as opposed to probabilistic
>> (inherently nondeterministic), see above. If you want to take me
>> too literally, of course, when there's the only event source (at
>> possessions of both differently prioritized channels) for the
>> event handler, it's still deterministic, since this source has
>> a full control of the requests (until they are running fully
>> asynchronously), but it's gone when there are independent sources.
> 
> Current loop is as deterministic as glib implementation. It's just
> different behavior.

I don't see how you come to this conclusion.

GLib makes a promise to serve higher priority events before proceeding
with lower priority ones.  Scheduling plan is a function of incoming
events (and their priorities) only.  If handling one high prio event
generates another one, it is deterministically ordered before any
events to handle that are of lower priority.

Libqb doesn't make such promise, rather it interleaves the priority
levels (assuming there's always some more events to handle) per some
kind of proportional system.   Scheduling plan is a function of
incoming events, and some hidden state that determines which step in
said proportional system is to be followed.  If handling one high prio
event generated another one, it is deterministically ordered within
events of the same priority (FIFO, I hope), but non-deterministically
ordered with respect to the lower prio ones (that moreover can
generated some other event that can moreover be handled prior to said
just queued high prio one, depending on said hidden state!).

>>> Also are you sure that your advice is really true for all (or at
>>> least most of the) "real-world" systems?
>>
>> I am actually not aware of systems where probabilistic approach to
>> priorities would get used (apart 
> 
> So no hard data

It's a straw man -- will you give me the budget for this en masse,
quantitative research?  Or what hard data are you after?

It's actually very simple, you'll achive the very same effect if you
offload any priority considerations out to event handling callbacks
themselves, each callback having a static counter i incremented and
modulo-checked per some factor:

- high priority: !(i % 1)
- med priority: !(i % 2)
- low priority: !(i % 3)

and you get the same functional behaviour as to what libqb offers.

I am yet to see a fairness balancing like that, it may exist,
but yeah, hard data would be handy.

On the other hand, it's quite common to see something like (pseudo-code):

	while not terminate:
		ret = poll(fds, nfs, 500)
		if ret == -1:
			if errno == EINTR:
				continue
			else:
				return -errno
		elif ret == 0:
			continue
		for pf in prioritized_fds:
			new_round = False
			for f in fds:
				if f.fd == pf.fd and f.events & f.revents:
					pf.handler()
					new_round = True
					break
			if new_round:
				break

>> from libqb). Hand-written loop management is also typically based
>> on implicit, strict priorities (statically 
> 
> Do you have any hard data?

I'd get them for you if time capacity only allowed...

To flip it, do you have any counter-examples?

>> hard-coded order of handling the particular events). The experience
>> (see the unit test) tells that GLib behaves in what I refer to as
>> deterministic manner. That's somewhat more intuitive, yet more
>> demanding 
> 
> It's easier to describe glib behavior, but I wouldn't call it more
> deterministic

See above.

>> when it comes to the correct usage (requiring
>> starvation/deadlock-avoiding design considerations). It's basically
>> "process niceness vs. real-time scheduling" analogy. Libqb favours
>> the former (without 
> 
> Kind of right analogy and maybe nice to have it in description,.

Ok, let me amend that comment.

> declaring that clearly, until the commit in question), GLib et al.
> the latter. Any suggestions how to word this better while still
> guiding the libqb users towards correct expectations welcome.
> 
> Because chrissie is on it I will not add my description but idea is
> to really describe how current implementation works and what may be
> drawbacks. That's it. No need to describe feeling and/or give advice
> without any real evidence.

It's basically a friendly advice to the fellow programmer willing to
use the facilities of libqb -- I am thinking it about what I'd make
clear to insight-less me if I were about to use qbloop.h.

I think that's the best to see it, incl. giving pointers to what
to use when the other, more intuitive (since "priority" is really
established as urgency in the field, since the age of IRQs if not
much older) behaviour is actually requested.

-- 
Jan (Poki)

jfriesse:
I'm going to ignore previous comment because it's just complete non-sense and waste of time so I'm commenting just result.

Maybe it's just me, but I find really really hard to read/follow the description and would be rather more happy with single line as proposed by chrissie.

jnpkrn:
This whole debate is mostly wasteful.
It's pretty cheap to kick down the detailed justifications.

You are not the primary audience of that guidance, newcomers to libqb
are, afterall.  They may use `qbloop.h` interface in contexts where
exact approach towards priority handling doesn't matter at all, since
they are not going to leverage that mechanism at all.  Or they may
rely on particular behavioral traits, we cannot guess here.
And nobody is obliged to read it, but if it prevents people shooting
themselves in the foot, great, and more context the better for these
reasonable individuals.

Also, making a connection to `qbipcs.h` explains the primary use case
of that interface, including touching on the ambiguity with rate
limiting (which just event/channel priority meaning in the sibling
module).

wferi:
Actually, this thread was very useful for me, as I always wondered what the priority levels of libqb really mean. So much thanks to @jnpkrn for displaying the details. Whether such details should be documented depends on the guarantees provided by the library. Based on @chrissie-c's comment it looks like the main point is the _lack_ of such guarantees. Which is valid behavior, but maybe not too useful, as it's very simple to implement in the application code anyway. Avoidance of starvation of any level is an interesting guarantee on the other hand, but that's again simple to achieve by using a single priority. So it isn't very clear to me what the priority system was designed for. Would it make sense to explain that in the documentation, perhaps?

jnpkrn:
[distributed systems sometimes fail, previous two attempts of
response-via-email didn't make it for unknown reasons, at worst,
there will be some dupes "eventually"]

On 30/05/19 06:46 -0700, wferi wrote:
> So it isn't very clear to me what the priority system was designed                                                                
> for.                                                                                                                              

I am hesitant to take a role of SW archeologist on me, since it can
be quite exhausting, but my guess is that it has something to do with
with "rate limitting" on the IPC server side...

> Would it make sense to explain that in the documentation,                                                                         
> perhaps?                                                                                                                          

...as already (weakly, for I didn't get to investigate the nitty-gritty
as pointed out) implicated in the current version of that comment.

(Yep, it all has something to do with the problem of knowledge
retention.  Another reason to avoid rather obvious one-liners
and to elaborate the context some more.)

jnpkrn:
[distributed systems sometimes fail, previous attempt of
response-via-email didn't make it for uknown reasons, at worst,
there will be a dupe "eventually"]

On 30/05/19 06:46 -0700, wferi wrote:
> So it isn't very clear to me what the priority system was designed
> for.

I am hesitant to take a role of SW archeologist on me, since it can
be quite exhausting, but my guess is that it has something to do with
with "rate limitting" on the IPC server side...

> Would it make sense to explain that in the documentation,
> perhaps?

...as already (weakly, for I didn't get to investigate the nitty-gritty
as pointed out) implicated in the current version of that comment.

(Yep, it all has something to do with the problem of knowledge
retention.  Another reason to avoid rather obvious one-liners
and to elaborate the context some more.)

-- 
Jan (Poki)

jnpkrn:
On 30/05/19 06:46 -0700, wferi wrote:
> So it isn't very clear to me what the priority system was designed
> for.

I am hesitant to take a role of SW archeologist on me, since it can
be quite exhausting, but my guess is that it has something to do with
with "rate limitting" on the IPC server side...

> Would it make sense to explain that in the documentation,
> perhaps?

...as already (weakly, for I didn't get to investigate the nitty-gritty
as pointed out) implicated in the current version of that comment.

(Yep, it all has something to do with the problem of knowledge
retention.  Another reason to avoid rather obvious one-liners
and to elaborate the context some more.)

-- 
Jan (Poki)

chrissie-c:
I'm quite happy to take a detailed explanation of the poll priorities and their uses if you can do it it modern, understandable English.

jnpkrn:
Pulitzer prize qualities != usability/information value.

You should have rather watched for the latter, see
https://github.com/ClusterLabs/libqb/pull/342#issuecomment-473362396

kgaillot:
> Pulitzer prize qualities != usability/information value.

Comments have to add more to understanding than confusion. Wording, spelling, and grammar don't have to be perfect, but intent should be clear and concise. It's always difficult to encapsulate our own thought processes in a way that others can clearly understand; that's why getting feedback from them is a helpful signal whether we are or aren't accomplishing that.

I would put the comment with the qb_loop_priority enum rather than the file block. Maybe:

    When multiple events are waiting to be handled, they will be processed proportionately
    to their priority -- that is, every event will eventually be processed, even if higher-priority
    events are continuously arriving.

It's not essential, though occasionally helpful, to contrast with other common approaches. I don't think it's necessary here, but I wouldn't oppose it, something along the lines of:

    By contrast, some other implementations (such as GLib's mainloop) always process
    higher-priority events first, requiring the caller to consider and handle deadlocks. The
    difference is similar to nice(1) priorities versus real-time scheduling of processes.

I disagree that proportionate priorities are somehow in conflict with what's "traditionally established in the field". nice(1) is a clear example of a very traditional, very established usage in the field.

It's never reasonable to disparage the current design simply because you prefer another. A proportional system does not provide a "false sense of correctness". Every design has advantages and disadvantages, and the disadvantages don't make them horrible, scandalous decisions that need to be decried from the very code that implements them. Also, a "sense of correctness" is not the highest goal of software development, solving real-world problems is. Formalized systems are ways to simplify the complexity of reality for better decision-making; they are not superior or valuable in and of themselves, other than as a sort of mental game.

jnpkrn:
So much energy that would rather be invested in code and tests.

[
Speaking of that, I rehashed the unit tests so that libqb and GLib
are clearly comparable so that the crucial difference is undisputable.
]

But I see it's tempting to comment on something with seemingly low
barrier to entry, such as attacking an attempt to show possible traps
to the fellow programmers -- very similar kind of traps pacemaker and
libqb itself fell into independently (and which are fixed in PRs by
now).

Interestingly, you still don't seem to be groking how tricky this can
be, and hence cannot even see how worthy the advises I put to the
file-scope level comment are :-/

They belong here, since, as mentioned several times, in generic
messaging and signaling in the field, it's established that highest
priority (or urgency) is what's in non-fuzzy use cases picked amongst
the available parallel candidates, since that's the most truthful
following of the intention of whoever/whatever assigned such priority
-- "if you want to perform an immediate shutdown, there's nothing of
a bigger priority".  Or critical disaster limiting actions.
In a system with such clearly given priorities, it's rather simple to
embed additional mechanisms for some relative tweaks for individuals
belonging to the same priority level.  Definitely easier than the other
way around -- how would you achieve big power based only on weak
primitives without out-of-band shortcuts (again refering to the analogy
of niceness vs. real-time scheduling, the former is just a small deal
once the big picture scheduling has been done based on hard constraints).
So those having experience and natural expectations with such a behaviour
shall really be told the truth from the get go.  And those wanting to
flip the implementation shall be reminded of the risks of paradigm
changes.

And that's exactly what's being done here!

> I disagree that proportionate priorities are somehow in conflict with
> what's "traditionally established in the field".

That's how most of the HW works, whether you like it or not.
And massively deployed SW such as likewise mentioned GLib agree,
and they are really not expected to document "don't be surprised
whenever low prio event is never interleaved with handling of the
sequence of high prio ones", since that's really a mainstream vision,
that's the power put at the hands of the implementors by such.

> nice(1) is a clear example of a very traditional, very established
> usage in the field.

Yeah, that's what I explicitly mentioned as an counterexample (to make
the potential library user better accustomed to the kind of a difference
that's going on here), but that's just a weak fine-tuning of the
important scheduling plan arrangement done before that -- which is what
does matter and is this "traditionally established in the field".

> It's never reasonable to disparage the current design simply because you
> prefer another.

Now, how do you come to this conclusion?  Former iterations could appear
like that, so I changed the possibly controversial parts.

> A proportional system does not provide a "false sense of correctness".

Of course it does, in the exact sense that burnt us in pacemaker, since
there's this freedom of a swap, which is exactly what was done.

What could be change is to put the advices regarding "swap problem"
into `qbipcs.h`, since that's where it is imminent the most.
Detached usage of "loop" module is less likely be an universally
changeable component is other SW.

> ... horrible, scandalous decisions ...

You picture your emotions here.

> Also, a "sense of correctness" is not the highest goal of software
> development, solving real-world problems is.

Again, some projections, we don't know who may want to use libqb for
what.  You put words into my mouth that I didn't say, do you enjoy
this play?

Nonetheless, for HA, correctness is quite an important thing, so one
more reason to be explicit rather than implicit.

jfriesse:
@kgaillot Thank you for your very nicely written constructive comment. It's basically what I was trying to explain to @jnpkrn but without too much success (but may be my bad)

kgaillot:
> But I see it's tempting to comment on something with seemingly low
> barrier to entry, such as attacking an attempt to show possible traps

This isn't an adversarial process, rather a way to iterate toward a solution that best meets different people's different goals and assessments.

> Interestingly, you still don't seem to be groking how tricky this can
> be, and hence cannot even see how worthy the advises I put to the
> file-scope level comment are :-/

It's not that the advice couldn't help someone at some point, but rather it's questionable whether this is the proper place for it, and even more questionable whether the wording accomplishes the goal.

Rethinking, I think this sort of code exegesis would be better suited to the ClusterLabs wiki. I'd recommend a brief comment like the one I suggested, along with a "For details see https://wiki..."

The wiki page could start with your wording; it's easier to iterate on a wiki and keeps the code uncluttered.

jnpkrn:
On 03/06/19 07:22 -0700, Ken Gaillot wrote:
>> But I see it's tempting to comment on something with seemingly low
>> barrier to entry, such as attacking an attempt to show possible
>> traps
> 
> This isn't an adversarial process, rather a way to iterate toward
> a solution that best meets different people's different goals and
> assessments.

I've explained clearly which optics I apply for that comment,
and think it makes a good job of pointing surprises, pros
and cons out.

>> Interestingly, you still don't seem to be groking how tricky this can
>> be, and hence cannot even see how worthy the advises I put to the
>> file-scope level comment are :-/
> 
> It's not that the advice couldn't help someone at some point, but
> rather it's questionable whether this is the proper place for it,
> and even more questionable whether the wording accomplishes the
> goal.
> 
> Rethinking, I think this sort of code exegesis would be better
> suited to the ClusterLabs wiki. I'd recommend a brief comment like
> the one I suggested, along with a "For details see https://wiki..."

That's a bad idea, library documentations shall rather be
self-contained (basically everything starting from kernel level is,
except for less targeted/more noisy cross-component FAQs, HOWTOs,
and, derived from that, essays, blog posts etc.).

With error messages where there's not enough room for details, that's
another story (having pacemaker in mind ... but even for that, it'd
be good to have text document installed along so the self-containment
can be applied also there).

> The wiki page could start with your wording; it's easier to iterate
> on a wiki and keeps the code uncluttered.

See above, moreover, the API headers are detached from the actual code
already, so the audience gets met in right designated place.

* * *

Btw. I refreshed the unit tests part of PR, you can observe the clear
difference in GLib's behaviour compared to libqb's native loop when
you override

```
global_loop_prio = QB_LOOP_MED
```

`test_ipc_dispatch_*_glib_prio_deadlock_provoke` functions.

-- 
Jan (Poki)

chrissie-c:
Thanks Ken, that wording looks good to me.

@poki I appreciate it's a very difficult thing to do a unit test for but is there a way of doing it that isn't so intrusive, it seems we're changing a lot of things in existing tests, which doesn't seem good to me. Also, why have the timeouts for existing tests all being changed?

jnpkrn:
On 03/06/19 08:12 -0700, Chrissie Caulfield wrote:
> I appreciate it's a very difficult thing to do a unit test for but
> is there a way of doing it that isn't so intrusive, it seems we're
> changing a lot of things in existing tests, which doesn't seem good
> to me.

All those things are of negligible impact on the things under test,
while bringing new qualities to how they are approached (more helpful
logging, greater speed, slightly less redundancy code-wise).

> Also, why have the timeouts for existing tests all being
> changed?

I thought this would be apparent ... since avoidable (with
counterbalance that I added just to make it sound, even it
could make do also without for how the connection attempt
loops is constructed) "sleep(1)" invocations were dropped,
it actually shifts the timing expectations for such affected
tests -- there's no spurious lowering of the bar
even if it can look like that from the first sight.

Whenever any issues are spotted, timeouts and/or the arrangement
can be changed, as always, but no negative effect is expected
(nor observed with casual testing), really -- I am now talking
about the existing tests.

Regarding "test_ipc_dispatch_shm_native_prio_deadlock_provoke"
failing, I've seen that as well, but it went away after the
"fix", I think I'll take a closer look.

-- 
Jan (Poki)

jnpkrn:
> Regarding "test_ipc_dispatch_shm_native_prio_deadlock_provoke"
> failing, I've seen that as well, but it went away after the "fix",
> I think I'll take a closer look. 

Fixed, looks good to me now.

chrissie-c:
Can we have Ken's text in the comment please?

jnpkrn:
They are not representative as to the tricky nature I'd like to point
out -- and which likely caused the problem I fixed in the first place.

Is the current form a blocker in any place?  If so, in which in
particular?

Point of arbiters is to judge fairly, not with the hive mind of the
clique they happen to be in.  I'd appreciate if that was the case.

Apparently, the discussed matter was rather interesting for others to
learn about (thanks @wferi for your attendance here!), and I think the
current form discusses the possible shortcomings in about the right
depth to the random comer to the libqb's event loop.

include/qb/qbloop.h
```diff
@@ -36,6 +36,37 @@ extern "C" {
...
+ * vs. established "niceness" property only adjusting proportions of their
+ * interleaving, as is represented with this libqb's native loop management.
+ *
+ * Word of caution: developing the application using the native loop
+ * implementation first and exchanging it for the other based on this opposite
+ * paradigm later on may cause said unexpected blocking situations to emerge,
+ * since the former just used to provide a false sense of correctness there.
+ * The inverse direction is naturally smoother, since the code was likely
+ * well-balanced regarding prioritites already, but less overall performance
+ * can be expected if they were downright fine-tuned for the use case before.
```
kgaillot: While this is fairly simple to read, it adds virtually nothing of use to a developer.

Overall, virtually none of this is documentation; most of it commentary, discussion, and comparison with other designs. As such, it is much more like a blog post. Creating a wiki page for such lengthy discussions and then linking that from the API documentation is a very reasonable approach. The API can have the bare minimum to document what the API actually does, and discussion of the implications of the design can move at a separate pace, outside the code repository.

jnpkrn:
Tried to rehash and re-allocate the commentary to follow the suggestion
for less wordiness and more practical usability.  Also put the more, IMHO,
correct (out of all range of arguments in the discussion) stress on why
some implementors may come short and will need to use another event
loop implementation ... CS [therefore IT, when done right] is in essence
based on problem reducibility, and more powerful means are more likely
to be easily reducible to weaker power requiring solutions, but not the
other way around (see, for instance, it took ages for yum to adopt the
proper solution to the "problem of satisfiability", for which there are
stock solutions, called SAT solvers -- all the struggle prior to that
could be avoided if this problem reducibility gist was duly followed).

jnpkrn:
(typo fix)

chrissie-c:
TBH I'm unhappy you pulled this change without my approval. I'm still not happy with the comment in qbloop.h and I really don't see the need for the glib code in the test suite. Yes, there are a lot of good things in the check_ipc.c patch here (for which thank you), but they really should have gone in a separate PR rather than bundled with this one.

jnpkrn:
On 12/06/19 08:07 -0700, Chrissie Caulfield wrote:
> TBH I'm unhappy you pulled this change without my approval.

It was OK'd by Ken without further responses.

> I'm still not happy with the comment in qbloop.h and I really don't
> see the need for the glib code in the test suite.

To demonstrate the problem (and fix thereof), as is customary on the
project, and what will hopefully be followed in the future, also to
retain the clarity of the encountered problems in the past and about
the desired behaviours, in a reproducible manner.

That's why unit tests are important, and if they only softly ask
for a new dependency (and without it no such tests can actually
exist), there's hardly any conflict.

> Yes, there are a lot of good things in the check_ipc.c patch
> here (for which thank you), but they really should have gone
> in a separate PR rather than bundled with this one.

That's the matter of causality.  Would the familiarity with the
context be that implicit as you seem to suggest, the problem would
not have been introduced with the respective refactoring in the
past (I touched on this several times in this discussion).

So it forms a logical unit addressing the particular problem space.
That's how patchsets are usually structured, since otherwise it's
easy to miss something for unnecessary postponing and/or to suffer
from open loops which were once a low hanging fruit for closing.

It's also economical, several minds are concentrated in particular
direction _once_ rather than repeatedly with more mental energy burnt.

Of course, there are trade-offs to be made, but this was only a single
commit for which detaching doesn't make much sense, given its
"on-topicness".

-- 
Poki

chrissie-c:
Ken is not libqb maintainer, much as I respect him. I am.

While I'm all in favour of good unit tests - of course. The ones we have are already annoyingly fragile, adding more complexity is not helpful. Especially adding a comparison with a different library is just asking for unexpected failures.

jnpkrn:
[sorry, my former response was in part based on misinterpreting
`check_ipc.c` as `qbloop.h`]

Re complexity:

That's why one of the commits introduced better traceability to which
role amongst multiple processes employed in the test is the source of
the message -- I've enjoyed much _fun_ without that originally, which
let me to add that.

I see the risk factors, new failures stemming there go onto my shoulders.

But was reasonably confident that the cut wasn't that massive, and total
time trimming was partially a side effect of adding new synchronization
that was needed between a particular phase of execution of what I call
`alphaclient` (running its priority events as a neverending stream) and
a regular client (that's in `GLib` case demonstrably starved out with
the former).  Since the mechanism was already devised, it was a low
hanging fruit to use it also for server-to-client synchronization,
for which purpose there was this `sleep(1)` formerly.

Frankly, I am constantly on the verge regarding complexity.
Best solution for that is to have rather clearly defined building blocks
without excessive stretch, and their frequent combination to achieve the
objectives, I think.  While I increased the complexity, it was an
amortized amount not proportional to the number of addressed code
locations, exactly for that reusing, and that better traceability was
also meant to counter that effect.  Next level, when it will seem
unbearable, will likely be to have an auxiliary module with
test-agnostic helpers (said building blocks) apart and well documented,
perhaps.  I don't know of better approach, sadly, apart from the worst
stance -- having it rot away for fears to touch anything.

chrissie-c:
I obviously didn't make myself clear.

While the initial patch was great, and the general updates to check_ipc.c are welcomed, there are things I do NOT want in libqb and think you've overstepped the mark in committing them without approval.

- The glib code has no place here. It's adding pointless complexity to the test suite for almost no perceivable value - it needs removing.
- The comment in qbloop.h is borderline incomprehensible and needs fixing. Either remove it completely or replace it with my shortened suggestion: "The priorities passed to the libqb loop functions are advisory and provide no guarantees as to the order the callbacks will be invoked."

These comments also apply to your patch (also committed without my approval) to the version_1 branch.

Thank you.

jnpkrn:
On 17/06/19 08:32 -0700, Chrissie Caulfield wrote:
> I obviously didn't make myself clear.
> 
> While the initial patch was great, and the general updates to
> check_ipc.c are welcomed, there are things I do NOT want in libqb
> and think you've overstepped the mark in committing them without
> approval.
> 
> - The glib code has no place here. It's adding pointless complexity
> to the test suite for almost no perceivable value - it needs
> removing.

It was explained multiple times that requiring such a cut is
undeserved.

Let me explain from another perspective.

Libqb IPC servers clearly require integration with some arbitrary
event loop to which it can be glued with three basic callbacks.

From examples, it's apparent that GLib was one such target for a long
long time -- configure.ac had a check for the presence of that
library (and don't forget, it's all optional, no GLib around is
entirely fine, examples won't get that support compiled in and the
respective unit tests will get skipped, and that's it).  Beyond that,
likely to make libqb fully self-contained (see my note in the previous
discussion wrt. knowledge retention), also a built-in event loop is
delivered, with rather different characteristics.

Pulling any other event loop external implementation into the unit
tests could be seen as abrasive, I agree.   In case of GLib, I cannot
say that, and such tests clearly demonstrate two things actually:

1. that there was indeed a bug -- not ever exposable with said
   native event loop, so for the purpose of regressions checking,
   it makes a tonne of sense to employ said once (ages ago)
   accustomed integration library (mind that there can be no such
   meaningful test otherwise), and validation of fix thereof
   (commits were purposefully ordered to make this transition
   from red to green observable!)

2. with a slight modification of the unit test (as described in the
   comment), it can be demonstrated clearly that GLib event loop
   works on the stonger basis than the native loop does, so it's
   rather trivial to verify the claims newly added to the
   documentation (see below), which in turn (again, as mentioned
   multiple times) were shown as rather critical in a sense that
   was there a full awareness of that, the original bug wouldn't,
   very likely, have been introduced (see defensive coding below)

> - The comment in qbloop.h is borderline incomprehensible and needs
> fixing.

Excuse me, but what is incomprehensible?
I think we had enough of this straw man.

Ken seemed to comprehend that well in the last iteration.

It's all for the sake of defensive coding.

Do you want the same mistakes to be repeated over, or you don't
follow that defensive coding also means mentoring others to avoid
falling into the traps?

> These comments also apply to your patch (also committed without my
> approval) to the version_1 branch.

I took it all for a closed chapter now, which I thought is
what it truly deserves -- rest in piece.  Problem is solved,
reproducibly regression-tested, and the wider space documented
to prevent unexpected surprises to us and others.
Exhumation should be well justified, IMHO.

(To be honest, I also wasn't ever consent to a removal of the linker
section efficient offloading of the log message book-keeping,
primarily for ABI destructing effect, and the removal wasn't even
fully concluded to at least retain some traits of the efficiency
gains that won't be possible if there's no enforcement on the
compile-time-constantness on some arguments that used to be enforced
like that in the past, see the Lars' posts on that topic on the list
-- perhaps something deserving more attention)

Speaking of what requires attention, in pacemaker, we are still
looking at being able to use open-socket-once approach to avoid race
conditions (still allowing some avoidable chances for DoS otherwise)
and perhaps to build a sort of a socket based activation out of that
(any guidance towards having such mechanism workable with libqb
welcome): https://github.com/ClusterLabs/libqb/issues/325

-- 
Jan (Poki)

PR #354, opened 2019-06-11
[version_1] IPC: server temporary channel priority loss

jnpkrn:
1.x backport of #352.


jnpkrn:
With Fedora build, I've noticed that `qb_ipcs_rate_limit` needs to be
outside the `#ifdef HAVE_GLIB` conditional.  Will send an update against
master as well.

PR #356, opened 2019-06-14
Faster IPC tests on *BSD

jnpkrn:
[turning the on-project branch into a regular PR so the discussion can
be had at all, that being said, would be entirely fine with old good
ML driven development, either using developers@cl.o or a new dedicated
list if PRs are not seen that viable]

This tries to nail the problem I added a cursory `/* ? */` comment at
recently.

Is it really the matter of *BSD being slow, or is it a systemic problem
of libqb (or the test code) exercising the something in a not-optimized-for
fashion?  I mean, VMs are of about the same speed, so it's natural
the differences amongst OSes will be minimal, but that apparently is not
the case.

Lowering the number of iteration is a big hammer to use when we are
out of ideas, I think.  Is it the case already?

### DIFF
--- a/tests/check_ipc.c
+++ b/tests/check_ipc.c
@@ -611,6 +611,13 @@ s1_connection_closed(qb_ipcs_connection_t *c)
 	if (multiple_connections) {
 		return 0;
 	}
+	/* Stop the connection being freed when we call qb_ipcs_disconnect()
+	   in the callback */
+	if (disconnect_after_created == QB_TRUE) {
+		disconnect_after_created = QB_FALSE;
+		return 1;
+	}
+
 	qb_enter();
 	qb_leave();
 	return 0;
--- a/tests/check_ipc.c
+++ b/tests/check_ipc.c
@@ -44,6 +44,13 @@
 #include "_failure_injection.h"
 #endif
 
+/* The stress test takes over an hour on FreeBSD otherwise */
+#if defined(QB_BSD)
+#define NUM_STRESS_CONNECTIONS 7000
+#else
+#define NUM_STRESS_CONNECTIONS 70000
+#endif
+
 static char ipc_name[256];
 
 #define DEFAULT_MAX_MSG_SIZE (8192*16)
@@ -1414,7 +1421,7 @@ test_ipc_stress_connections(void)
 	pid = run_function_in_new_process("server", run_ipc_server, NULL);
 	fail_if(pid == -1);
 
-	for (connections = 1; connections < 70000; connections++) {
+	for (connections = 1; connections < NUM_STRESS_CONNECTIONS; connections++) {
 		if (conn) {
 			qb_ipcc_disconnect(conn);
 			conn = NULL;


fabbione:
@jnpkrn in our CI, BSD is on pure baremetal and linux on VMs. Yet we see BSD failing with timeouts due to the time it takes to run that many iterations.

I have 2 questions:

1) is it really necessary to run 70000 iterations to verify the functionality? can we be happy to have less iterations and feel safe that the code is stable?
2) do we have a BSD expert in the team that could nail it down to lack of optimizations in libqb?

We know the code works but it is slow, that is better than faster-but-broken code ;)

I agree ideally this should not be a problem in the first place, but we don't have resources or knowledge to go that deep into BSD.

As a temporary measure (read: let's at least have BSD run the full test suite each build) we should reduce the iterations, while perhaps we could contact some BSD experts to help us optimize the code.

Make sense?

jnpkrn:
Thanks, Fabio, this answers the very basic thing that wasn't answered
before -- and I think rather undeservedly, and that's where I wanted to
direct the discussion.

It's totally OK to say, sorry, we don't have capacity to pinpoint the
BSD sloweness we are perfectly aware of, but let's be frank directly
in the code comment, e.g.

```
/* XXX: polling seems to be very slow with FreeBSD, cut down the iterations
        for now, shall eventually be investigated and resolved if possible */
```

Heck, I'm not even sure which BSD in particular you are talking about,
so correct the above claim if suitable.

In other words, to make a rather private experience shared with everyone
(so they can perhaps look into that would they want to scratch that itch,
which is definitely easier if everyone is on the same page -- currently,
I don't think this is the case, for a lack of context detail).

fabbione:
On 17/06/2019 17.45, Jan Pokorný wrote:

> 
> Heck, I'm not even sure which BSD in particular you are talking about,
> so correct the above claim if suitable.

FreeBSD12 and FreeBSD-devel (13). You have access to CI like everybody
else to look at logs etc.

Fabio

jnpkrn:
On 17/06/19 20:21 -0700, Fabio M. Di Nitto wrote:
> On 17/06/2019 17.45, Jan Pokorný wrote:
> 
>> 
>> Heck, I'm not even sure which BSD in particular you are talking about,
>> so correct the above claim if suitable.
> 
> FreeBSD12 and FreeBSD-devel (13). You have access to CI like
> everybody else to look at logs etc.

That's nice if I can immediately check that (not that I would want
to hunt that down over the rather unfriendly UI, it shall be the
patchset author's liability to describe what doesn't fit in the
current arrangement accurately, to justify the need for a change),
but don't forget that the only source there may be to reason about
the particular commits (git bisect and friends) is just the plain
git history + code itself.  That's one of the reasons behind a practise
of all-you-need-to-know details being customarily dumped into the
respective commits and/or code comments [*].

There may not be any such CI infra several years down the road,
and if so, how likely is it that logs from about the commit's time
will be recoverable, and if so, readily understandable (IT archeology
is more work than it may seem from a superficial look, definitely
more work than dedicating several seconds to add a comment akin to
what I suggested).

Thanks for considering that aspect.

[*] that was also discussed some years back, see point 1. at
    https://lists.clusterlabs.org/pipermail/developers/2018-January/001958.html

-- 
Jan (Poki)

chrissie-c:
It is NOT appreciated that you take branches that I am using for experimentation and turning them in PRs without asking me. When I am ready for this to be a PR I will create one

jnpkrn:
On 19/06/19 05:10 -0700, Chrissie Caulfield wrote:
> It is NOT appreciated that you take branches that I am using for
> experimentation and turning them in PRs without asking me. When I am
> ready for this to be a PR I will create one

NP, but please, be so kind to do that in your local "fork", then.

It is unclear why the authorative repo would be used for that work
without even separating feature/WIP branches clearly.

Also, where do you expect to receive early feedback like the above
if that work is not announced in any other way?  What I don't want
to see is those commits appearning in important branches all of
a sudden, and here, I just noticed that very new branch accidentally,
without any further pointers.

-- 
Jan (Poki)

chrissie-c:
It's in that repo because it's the easiest way to get it into the CI system for testing (which is where the problem manifests itself).

jnpkrn:
Fair enough, just that we could at least have some clear prefix for work
in progress branches that are unavoidable for such reasons, like `wip-`.

It's than clear on several assumptions, like that such branch is not
expected to stay around forever.  And wouldn't raise reason seeking
questions.

PR #357, opened 2019-06-19
ipc: Fix named socket unlink on FreeBSD

jfriesse:
Terminating NUL on FreeBSD is not part of the sun_path.
Add it to use sun_path as a parameter of unlink.

### DIFF
--- a/lib/ipc_setup.c
+++ b/lib/ipc_setup.c
@@ -600,6 +600,13 @@ qb_ipcs_us_withdraw(struct qb_ipcs_service * s)
 		socklen_t socklen = sizeof(sockname);
 		if ((getsockname(s->server_sock, (struct sockaddr *)&sockname, &socklen) == 0) &&
 		    sockname.sun_family == AF_UNIX) {
+#ifdef HAVE_STRUCT_SOCKADDR_UN_SUN_LEN
+			/*
+			 * Terminating NUL on FreeBSD is not part of the sun_path.
+			 * Add it to use sun_path as a parameter of unlink
+			 */
+			sockname.sun_path[sockname.sun_len - offsetof(struct sockaddr_un, sun_path)] = '\0';
+#endif
 			unlink(sockname.sun_path);
 		}
 	}


lib/ipc_setup.c
```diff
@@ -600,6 +600,13 @@ qb_ipcs_us_withdraw(struct qb_ipcs_service * s)
 		socklen_t socklen = sizeof(sockname);
 		if ((getsockname(s->server_sock, (struct sockaddr *)&sockname, &socklen) == 0) &&
 		    sockname.sun_family == AF_UNIX) {
+#ifdef HAVE_STRUCT_SOCKADDR_UN_SUN_LEN
+			/*
+			 * Terminating NUL on FreeBSD is not part of the sun_path.
+			 * Add it to use sun_path as a parameter of unlink
+			 */
+			sockname.sun_path[sockname.sun_len - offsetof(struct sockaddr_un, sun_path)] = '\0';
```
jnpkrn: Nice spot.

If I understand it correctly, the problem is that we don't follow
this guidance from FreeBSD header:

```
/*
 * Definitions for UNIX IPC domain.
 */
struct sockaddr_un {
        unsigned char   sun_len;        /* sockaddr len including null */
        sa_family_t     sun_family;     /* AF_UNIX */
        char    sun_path[104];          /* path name (gag) */
};
```

since the calculation of that field (`sun_len`) on our libqb side
uses `SUN_LEN` macro, which does _not_ include the "null" (it uses
just `strlen` as far as `sun_path` is concerned).

Shall we not solve that discrepancy instead?

It looks like Linux is either of another opinion, or tolerant to
both approaches.  POSIX won't help in this regard :-(

jfriesse: I've gave a try to `SUN_LEN(...) + 1` on all lines where `sun_len` is set instead of using patch in this PR and result is that problem reappeared. So I've added print of `sun_len` before `unlink` and length didn't changed. So I've (just for fun) set it to 100 on all places and ... surprise surprise ... original problem appeared again and print of `sun_len` was keep showing old value.

So my conclusion is, that sun_len is more or less ignored and set internally by libc/kernel/....? Maybe some better compatibility with Linux?

Anyway, proposed patch is probably only way how to get correct behavior when using `getsockname`.

jnpkrn: Interesting info, and rather surprising (regarding usability, the comment
is rather off then -- it's actually excluding null -- which may be downright harmful)

Then the only possible problem is off-by-one buffer overrun, which appear
already precluded by how we use `snprintf` to set the path.

ACK.

jfriesse: You are talking about FreeBSD header, right? If so, than yes, it's pretty off and FreeBSD behavior seems to be quite harmful. Unix man page (the source of my comment) is much more clear.

jnpkrn: Right (and contrary, took me a moment to parse "Unix" ~ `unix(4), or `unix(7)` with Linux)

chrissie-c:
Good catch, thank you.

PR #359, opened 2019-06-27
Improve speed of IPC tests

chrissie-c:
Reduce the number of iterations from 70000 to 5000. For what it's testing (multiple connections) the extra large number doesn't tell us anything significant, and just adds to the load on the CI and build systems - especially BSD but also the slower Linux systems. This has been a collegiate decision after discussion on IRC.

The EV_EOF message has been removed from the FreeBSD kqueue code - there is no equivalent of this on the epoll side and, again, it just slowed things down.

There's also a user-after-free bugfix to test_ipc_disconnect_after_created(), which currently only affects BSD but in theory could trigger anywhere.

### DIFF
--- a/lib/loop_poll_kqueue.c
+++ b/lib/loop_poll_kqueue.c
@@ -153,8 +153,6 @@ retry_poll:
 			revents |= POLLERR;
 		}
 		if (events[i].flags & EV_EOF) {
-			qb_util_log(LOG_INFO,
-				    "got EV_EOF on fd %d.", pe->ufd.fd);
 			revents |= POLLHUP;
 		}
 		if (events[i].filter == EVFILT_READ) {
--- a/tests/check_ipc.c
+++ b/tests/check_ipc.c
@@ -44,6 +44,8 @@
 #include "_failure_injection.h"
 #endif
 
+#define NUM_STRESS_CONNECTIONS 5000
+
 static char ipc_name[256];
 
 #define DEFAULT_MAX_MSG_SIZE (8192*16)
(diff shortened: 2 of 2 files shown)


fabbione:
looks good to me, simple and clean

chrissie-c:
Thanks.

jnpkrn:
Could you hold the PR open at least for a day, please?

That's somewhat expected in this workflow, and review by
someone with blind spots in some areas won't help much.
In other words, @chrissie-c, you fulfilled exactly those worries
I was explicit about in the very related and purposefully created
PR #356 (making it fully justified in retrospect).
Why so blatantly crossing such lines if better code base
cannot be such reason?

Merely on technical grounds, in particular, `disconnect_after_created`
variable seems handled at the wrong level of abstraction -- it shall
be in the same context it was set initially, to abid the logic of
"context enter", "do something", "context leave".  That very
global is not alone and we'd be soon mad when we handled restoring
to defaults at some obscure places, which appears what's happening
here.

Then, what's the exact point of returning `1` in `s1_connection_closed`
on that occasion of neutralizing the global?  Looks pretty superfluous
to me, but lacking further details on why this has been done like that.

If there're no particular reasons, please, fix said misplacement
of the resetting of a global, it should really be at same level and
scope where initial "tainting" is, so as to preserve and maintain code
sanity throughout.

* * *

Regarding the number of iterations, it'd be pointless to argue with
you, but I believe the instincts of the original authors it was well
justified to afford more time (tens of seconds, normally, the savings
should have been part of the commit message, really) to get more
reliable (rather bit longer running) stress testing.

PR #366, opened 2019-10-09
[build] add --with-sanitizers= option for sanitizer builds

fabbione:
those options are stricly meant for runtime debugging purposes.
do NOT use in production.

check gcc man pages on how to use ASAN and UBSAN.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -534,6 +534,12 @@ AC_ARG_ENABLE([fatal-warnings],
 AC_ARG_ENABLE([debug],
   [AS_HELP_STRING([--enable-debug],[enable debug build])])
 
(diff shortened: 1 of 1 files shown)


ThomasLamprecht:
I used such options quite a bit and the integrations looks OK, so FWIW:
Reviewed-by: Thomas Lamprecht <t.lamprecht@proxmox.com>
Tested-by: Thomas Lamprecht <t.lamprecht@proxmox.com>

Note for ASAN: as the tests seldom care about freeing allocated memory, there are tons of "false positive" (i.e., no one cares for that in short lived testing programs) so one needs to run the check target with something like `export ASAN_OPTIONS="detect_leaks=0"` to filter those out.

PR #368, opened 2019-11-07
ipc: Always initialize response struct

jfriesse:
Response structure was not initialized completely,
when mkdtemp/chown failed, server was not accepting connection yet or
connect failed for some reason.

This is not an issue, but valgrind reports this
as a problem so it is easy to miss real problem then.

Solution is to initialize response before it is used.


kgaillot:
Code looks good to me. Travis CI's "make rpm" is failing with "error: line 9: Empty tag: Version:" which I am sure is unrelated but needs investigation.

chrissie-c:
Looks good to me thanks. The 'make rpm' failure only affects PR builds for some reason. I've looked into it but haven't worked out why

PR #371, opened 2019-12-06
tests: bump ipc timeout a bit to deal with CI limited resources

fabbione:


fabbione:
there is more than just timeout in the game, we are investigating the problem

fabbione:
problem is a buffer overrun in check_ipc.c. Proper PR will follow

PR #372, opened 2019-12-06
tests: Shorten deadlock test names

chrissie-c:
On newer Fedora systems that can have 32 bit PIDs, these long test
names can get truncated in the libqb internal buffers and thus break the
tests, so I've shortened the names.


fabbione:
ack

PR #378, opened 2020-01-08
build: allow for possible v1 branch continuity by generous SONAME offset

jnpkrn:
The main and the most ABI-touching thing for the envisioned 2.0 branch
is the usage of the linker-build-time allocated callsite info, avoiding
the non-economic evaluations and, under some circumstances dangerous,
heap allocations in the run-time.

Considering that v1.9.0 release (libqb.so.20) was expressly marked as
tech-preview[1,2] (hence something that shall not make it to production
use), there should be no harm for master branch (that is headed towards
2.0 and beyond) to receive noticable SONAME bump (libqb.so.100) so as to

- leave enough of space for a possible v1-compatible branch evolution
  (for use cases where recompile-everything is a no-go).
  in particular, with resuming with libqb.so.30, there would
  be a room for 99-33 = 63 add-new-drop-nothing compatible
  changes for that branch (which is more than plentiful)

- indicate some big change is going on more clearly towards client space

This is supposed to be a reasonable trade-off solution that would still
leave enough wiggle space, and would represent responsible approach to the
development (like the original attempt to prevent ABI break in the first
place was), allowing for more than an enforced unanimity (rather
antagonistic in the free software realms).

[1] https://lists.clusterlabs.org/pipermail/users/2019-December/026690.html
[2] https://github.com/ClusterLabs/libqb/releases/tag/1.9.0

### DIFF
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -30,7 +30,7 @@ AM_CPPFLAGS             = -I$(top_builddir)/include -I$(top_srcdir)/include
 
 lib_LTLIBRARIES		= libqb.la
 
-libqb_la_LDFLAGS	= -version-info 20:0:0
+libqb_la_LDFLAGS	= -version-info 100:0:0
 
 source_to_lint		= util.c hdb.c ringbuffer.c ringbuffer_helper.c \
 			  array.c loop.c loop_poll.c loop_job.c \


chrissie-c:
I really don't think that's necessary. All the new features are going into the v2 (currently master) branch andv1 is just bug-fixes and maintenance.

jnpkrn:
On the contrary, is there any factual reason not to bump it like
that?

It makes the room, reserves space for the future uknowns.
Industry is full of examples of that, since we all rely on some
sort of enumerability, but sometimes wish for subdividing the
space between two items in the enumeration, and better not to
wish for that just in retrospect, incapable of any action that
would had to be made in the past.  If they get unused nonetheless,
no harm occurred, there's nothing to lose, unlike with that
total lack of opportunity in some unfoldings.

Now, luckily, we are capable to deliberately decide to keep
some bridges in place.  Sometimes, even maintenance means
some slight breaks.

Chrissie, let's do it, _please_.

Schedule-wise, it may actually (or should) come just about prior
to release, I think, as a conclusive "readiness" touch (perhaps
followed/interleaved with commit like the current HEAD/e830a68
that is to serve "baseline version" sourcing in case there are
no other means ... think "tar czpf libqb.tar.gz $(git ls-files)").

* * *

Also, I'll have some inputs on what's begging for v2.0 release.
Let's please coordinate around linked
https://github.com/ClusterLabs/libqb/projects/1.
This "release candidate" was rather sudden to me (OK, still an RC,
but you see that it's what spurred attention with some worries
2.0 will be just as sudden) and I think constructive approach
looks more like getting people in the same boat, free of
surprises.

chrissie-c:
On further consideration I agree with you. Although I still very much doubt the extra numbers will ever be needed it seems churlish to preclude the possibility.

Thank you.

The 2.0 release will not be done in a rush. I'm busy on other things at the moment and anyway, some of these features have been languishing in the master since early last year. I'm mainly trying to get people to try them for now.

chrissie-c:
pulled but GH didn't spot it.

jnpkrn:
Thanks.  The proven trick when dealing with commits manually
(that I encourage and applaud to here), is to state something like
this in the commit:

```
Closes: #378 
```

Already used in
https://github.com/ClusterLabs/libqb/commit/c1bca92c8915df6d7625661885e25fa9073a82e3

Feature details:
https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords

PR #380, opened 2020-02-06
tests: Run IPC with use-filesystem-sockets active

chrissie-c:
Provide an LD_PRELOAD library that simulates the presence
of /etc/libqb/use-filesystem-sockets so that we can test
that functionality without actually having the file on
the system and affecting everything else running on the
box.

NOT FOR MERGING YET. scripts need testing on other arches at least

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -113,11 +113,16 @@ endif
 bench_log_SOURCES = bench-log.c
 bench_log_LDADD = $(top_builddir)/lib/libqb.la
 
+lib_LTLIBRARIES = libstat_wrapper.la
+libstat_wrapper_la_SOURCES = libstat_wrapper.c
+libstat_wrapper_la_LIBADD = -ldl
+libdir= $(TESTDIR)
+
 if HAVE_CHECK
-EXTRA_DIST += start.test resources.test
+EXTRA_DIST += start.test resources.test ipc_sock.test
 EXTRA_DIST += blackbox-segfault.sh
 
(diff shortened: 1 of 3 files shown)


wferi:
Hi Chrissie, why doesn't introducing switching by some environmental variable solve this problem (in a less obscure way)?

chrissie-c:
@wferi I forget all the detailed discussion we had about this, but the filesystem file was the best solution. One argument against the variable I do remember is that the client and server MUST be the  same otherwise the IPC won't work. eg. if the server sets up an FS socket and the client tries an implicit socket. There's more to it than that (it was quite a long email trail) but that was one of the reasons. It's a very specific use-case for openstack and not really a general feature TBH.

wferi:
Yeah, I get the point about the filesystem being a handy system-wide switch (until you consider mount namespaces). I didn't mean to replace but to augment that, mostly to ease testing, just like this wrapper. However, the environment variable could even override the compile-time default socket directory, enabling more complex setups with several connected client/server groups without using the Linux-only mount namespace feature. (I tried but failed to find the past discussions about this before asking, sorry to rehash those.)

chrissie-c:
Yeah. As I said, this is for a specific use-case so it's mostly tailored for that. If you have another use-case that can use this feature in a different context then I'll take patches ;)

kgaillot:
Another advantage: The primary use case is having containers that use their host's network use filesystem sockets while the host simultaneously uses native sockets, so there is no chance of conflict. Using a file as a flag conveniently allows it to be built into the container image.

chrissie-c:
retest this please

chrissie-c:
retest this please

Issue #369, opened 2019-11-14
Bad owner of /dev/shm/qb-65059-65122-12-lJbC1D directory


radosroka:
Why is the owner group of the ```qb-65059-65122-12-lJbC1D``` set to rsroka when it should be ```root``` as it is(correctly) in files inside that directory. 

The USBGuard is setting this via ```qb_ipcs_connection_auth_set``` in Connection Accept fn.
```
qb_ipcs_connection_auth_set(conn, uid, 0, 0660);
```

```
[root@Axis shm]# ls
qb-65059-65122-12-lJbC1D
[root@Axis shm]# ls -la
total 0
drwxrwxrwt.  3 root   root     60 Nov 14 17:39 .
drwxr-xr-x. 23 root   root   4540 Nov 14 15:58 ..
drwxrwx---.  2 rsroka rsroka  160 Nov 14 15:36 qb-65059-65122-12-lJbC1D
[root@Axis shm]# ls -la qb-65059-65122-12-lJbC1D/
total 3120
drwxrwx---. 2 rsroka rsroka     160 Nov 14 15:36 .
drwxrwxrwt. 3 root   root        60 Nov 14 17:39 ..
-rw-rw----. 1 rsroka root   1052672 Nov 14 15:36 qb-event-usbguard-data
-rw-rw----. 1 rsroka root      8248 Nov 14 15:36 qb-event-usbguard-header
-rw-rw----. 1 rsroka root   1052672 Nov 14 15:36 qb-request-usbguard-data
-rw-rw----. 1 rsroka root      8252 Nov 14 15:36 qb-request-usbguard-header
-rw-rw----. 1 rsroka root   1052672 Nov 14 15:36 qb-response-usbguard-data
-rw-rw----. 1 rsroka root      8248 Nov 14 15:36 qb-response-usbguard-header
[root@Axis shm]# 
```

The USBGuard is running an IPC "server" as a root and usbguard-notifier or usbguard-applet-qt are running an IPC "client" under standard user. This permission inconsistency results in extending daemon capabilities with DAC_OVERRIDE which is not good at all.

PR #381, opened 2020-02-10
Set correct ownership if qb_ipcs_connection_auth_set() has been used

diabonas:
- The temporary directory that is used since 6a4067c1d1764d93d255eccecfd8bf9f43cb0b4d needs to be accessible by the group as well. This has been fixed on the [version_1](https://github.com/ClusterLabs/libqb/tree/version_1) branch (and in version 1.0.5) by 65d6fb37a28f3b41f2e73c49214a4c9e3533f15a, but not on master. Cherry-pick this fix (as well as another related fix to make the patch apply cleanly) for master.
- When [`qb_ipcs_connection_auth_set`](https://github.com/ClusterLabs/libqb/blob/6a4067c1d1764d93d255eccecfd8bf9f43cb0b4d/lib/ipcs.c#L938-L946) is used, the ownership of the temporary directory that is initially set in [`handle_new_connection`](https://github.com/ClusterLabs/libqb/blob/7f891f0069623edc0b2212bf559e70fa56ba5c12/lib/ipc_setup.c#L663) must be updated as well, so do this at the beginning of [`qb_ipcs_shm_connect`](https://github.com/ClusterLabs/libqb/blob/7f891f0069623edc0b2212bf559e70fa56ba5c12/lib/ipc_shm.c#L310).

This pull request fixes USBGuard/usbguard#289: when running USBGuard without the `CAP_DAC_OVERRIDE` capability (as done by the provided system service or by running `usbguard-daemon -C` to drop unnecessary capabilities), the temporary directory cannot be accessed because its group ownership is set to the user of the IPC client despite USBGuard using [`qb_ipcs_connection_auth_set(conn, uid, 0, 0660)`](https://github.com/USBGuard/usbguard/blob/50c99211d512b963a8bcf271e316b65b967ab585/src/Library/IPCServerPrivate.cpp#L261) to set the gid to root. As a result, all IPC communication with non-root users is broken.

Fixes: #369

### DIFF
--- a/lib/ipc_setup.c
+++ b/lib/ipc_setup.c
@@ -660,11 +660,12 @@ handle_new_connection(struct qb_ipcs_service *s,
 		res = -errno;
 		goto send_response;
 	}
-	res = chown(c->description, c->auth.uid, c->auth.gid);
-	if (res != 0) {
+	if (chmod(c->description, 0770)) {
 		res = -errno;
 		goto send_response;
 	}
+	/* chown can fail because we might not be root */
+	(void)chown(c->description, c->auth.uid, c->auth.gid);
 
 	/* We can't pass just a directory spec to the clients */
 	strncat(c->description,"/qb", CONNECTION_DESCRIPTION);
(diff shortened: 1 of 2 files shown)


chrissie-c:
Good catches. Thanks!

PR #382, opened 2020-02-10
[version_1] Set correct ownership if qb_ipcs_connection_auth_set() has been used

diabonas:
Backport #381 in case there is another 1.0.* release.


diabonas:
The CI failure comes from a [linting error](https://travis-ci.org/ClusterLabs/libqb/jobs/648347719#L1914) that appears to be unrelated to the proposed patch.

chrissie-c:
It looks good to me. Can you fix the split errors please? it's just a question of changing the line in lib/Makefile.am from " 2 code warnings" to " 3 code warnings", then I'll pull it.

Thanks!

diabonas:
Thanks, I didn't realise the linting error was due to the ignored return value of `chown()`, on my system `run_splint.sh` fails with completely unrelated error messages. I amended the commit and force-pushed - not quite sure why CI is failing this time, it passed in [my own Travis CI namespace](https://travis-ci.com/diabonas/libqb/builds/148589303)...

chrissie-c:
Thanks for that. The RPM build fail is 'normal' so that I'm not going to worry about!

PR #383, opened 2020-02-17
list: fix list handling for gcc10

chrissie-c:
DO NOT PULL - this patch generates warnings, but I want to test
it in the CI before sorting out the detail.


chrissie-c:
This is ready for review now.

chrissie-c:
Thanks for the review

PR #385, opened 2020-02-25
qblist: Retype ptr in qb_list_entry to char*

jfriesse:
This allows pointer arithmetics without issuing warning.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -601,6 +601,7 @@ WARNLIST="
 	suggest-attribute=format
 	property-attribute-mismatch
 	strict-prototypes
+	pointer-arith
 	write-strings
 	cast-align
 	bad-function-cast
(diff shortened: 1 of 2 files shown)


jfriesse:
@chrissie-c A bit more length comment. I was trying to find out why corosync build was failing only in CI and not when testing on my machines. The reason is, that CI is using non-system libqb. Gcc (and probably other compilers?) have exception for header files in system directories and all warnings are suppressed. So`gcc -I$PATH_TO_QB cpg.c` would fail as long as $PATH_TO_QB is not (for example) `/usr/include`, if it is, it would succeed.This is achievable for other directories by using-isystem flag, but honestly I don't see it as a best way to go.

The patch in kernel which changed (char *) to (void *) - https://github.com/torvalds/linux/commit/c7acec713d14c6ce8a20154f9dfda258d6bcad3b - was, AFAICT, not that much about (char *) but about `const`. We don't have that, so I think retype to (char *) should do the job just fine without any problems.

Another possibility would be to remove `Wpointer-arith` from all of our projects and have big warning in libqb docs about this behavior (so using system installed libqb is fine, but if it is outside of system directories it will produce warning). IMHO not omptimal solution.

chrissie-c:
Thanks for your investigations into this. It's good to know what was going on!

PR #387, opened 2020-03-09
doxygen2man: Add doxygen2man utility

chrissie-c:
This tool (currently) isn't used in libqb, but may be in the future.
It's here so it can be used by other cluster  packages without them having their own private copy. It's is currently used by knet (where this code comes from) and we intend to also use it in corosync.

I'm not claiming this is the ideal place for this utility, but having a totally separate main project for it seems unnecessary.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -39,7 +39,7 @@ ACLOCAL_AMFLAGS		= -I m4
 
 dist_doc_DATA		= COPYING INSTALL README.markdown
 
-SUBDIRS			= include lib docs tools tests examples
+SUBDIRS			= include lib docs tools tests examples doxygen2man
 
 doxygen:
 	$(MAKE) -C docs doxygen
--- a/configure.ac
+++ b/configure.ac
@@ -167,6 +167,9 @@ if test x"$have_glib" = xyes; then
 AC_DEFINE_UNQUOTED([HAVE_GLIB], [1], [We have glib])
 fi
 
+# For building doxygen2man - ONLY
+PKG_CHECK_MODULES([libxml_BUILD], [libxml-2.0])
+
 # Checks for header files.
 AC_HEADER_STDC
 AC_HEADER_SYS_WAIT
@@ -792,6 +795,7 @@ AC_CONFIG_FILES([Makefile
 		 tests/Makefile
 		 tests/test.conf
 		 examples/Makefile
+		 doxygen2man/Makefile
 		 docs/Makefile
 		 docs/common.dox
 		 docs/html.dox
--- a/doxygen2man/Doxyfile.example
+++ b/doxygen2man/Doxyfile.example
@@ -1,0 +1,19 @@
+#
+# Copyright (C) 2017-2020 Red Hat, Inc.  All rights reserved.
+#
+# Author: Fabio M. Di Nitto <fabbione@kronosnet.org>
+#         Christine Caulfield <ccaulfie@redhat.com>
+#
+# This software licensed under GPL-2.0+
+#
+# Example file for doxygen when using doxygen2man
+#
+PROJECT_NAME	       = libqb
+PROJECT_NUMBER         = 2.0
+INPUT                  = ../include/qb/
+XML_OUTPUT             = ./xml
+GENERATE_XML           = YES
+XML_PROGRAMLISTING     = NO
+AUTOLINK_SUPPORT       = NO
+GENERATE_HTML          = NO
+GENERATE_LATEX         = NO
--- a/doxygen2man/Makefile.am
+++ b/doxygen2man/Makefile.am
@@ -1,0 +1,20 @@
+#
+# Copyright (C) 2017-2019 Red Hat, Inc.  All rights reserved.
+#
+# Authors: Fabio M. Di Nitto <fabbione@kronosnet.org>
+#          Federico Simoncelli <fsimon@kronosnet.org>
+#
+# This software licensed under GPL-2.0+
+#
+
+MAINTAINERCLEANFILES	= Makefile.in
+
+bin_PROGRAMS	= doxygen2man
+
+all: $(PROGRAMS)
+
+doxygen2man_SOURCES = doxygen2man.c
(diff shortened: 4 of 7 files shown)


kgaillot:
I don't have an opinion one way or the other, but a separate project would be worthwhile if a different release cadence would be useful, or if this might take off among projects that don't need libqb.

FYI, I've been working on converting Pacemaker (non-API) documentation to sphinx. Sphinx has doxygen-like capabilities for Python, and is able to output python API documentation as man, html, etc. The "hawkmoth" sphinx extension extends this to C, but I haven't looked into it yet. Alternatively the "breathe" project takes doxygen output and converts it for use by sphinx. I like the idea of having one system for API and non-API documentation, so I might look into that, though I don't know when I'll have time for it.

chrissie-c:
We have considered this, but the tool is really in "it works, leave it" mode. Most people wanting documentation from doxygen these days want more than just man pages AFACIT,but we're low enough in the stack that they are our main documentation (for APIs at least).

The resulting package is deliberately not prefixed with 'libqb' just so we can spin it off if it becomes necessary!

fabbione:
retest this please

wferi:
Please reconsider creating a separate project for doxygen2man. Cross-building is such a pain if you need to run some of your code during the build... See all those `*_BUILD*` variables in Kronosnet. (It will bite if you ever want to use it in libqb.) Thanks.

fabbione:
@chrissie-c I am fine either ways, if we want to make it a separate project or not. We have all the infra in place for it. Just need some autotools magic here and there.

jfriesse:
+1 for ether way - separate project or libqb (actually, best place would be in doxygen itself where it really belongs to as one of the output options).

wenningerk:
@wferi: that is where cross-build-environments like scratchbox shine - even useful for basic testing

fabbione:
@wferi this version of doxygen2man would replace the one in knet and shared across different projects. To be honest, I understand your concerns about cross-compile, but we have the autotools snippets in knet already (just need to be moved over here) and doxygen2man is too small to make it a project on its own IMHO.

jfriesse:
@wferi We discussed this problem a bit on IRC, and there are basically two problems:
- If doxygen2xml is a separate package and libqb dependency is not removed and if it is used for libqb doc then there is a circular dependency
- If it is in libqb, then there is problem with cross-compilation

I think having circular dependency is probably worse of two evils, but you may have some idea how to solve it in a better way?

wenningerk:
In case of cross-building:

BuildRequires: qemu-user-static

as long as the target isn't too exotic ...

wferi:
Oh well, I forgot about the libqb dependency of doxygen2xml. As far as I know, it's only the hash implementation, but I've got no idea what it would take to remove the dependency.
* The circular dependency is rather unfortunate, but not the end of the world, because it affects the documentation generation only, libqb is functional without it. This means bootstrapping would require a (partial) rebuild of libqb after stage1 (or a separate arch-all build to generate the libqb-doc package).
* Moving the current cross-compilation support from Kronosnet into libqb won't be enough when libqb starts to use doxygen2xml, because double compilation will be needed then. But it's something to build on at least. I'm not sure I'd dare to recommend requiring qemu-user-static in a face-to-face meeting with the cross-bootstrap crew, though. :)

So yeah, neither option is particularly pretty, but both could be made to work. The easiest solution is to not use doxygen2xml in the libqb build, then neither problem arises regardless if it's a part of libqb or a separate project.

jfriesse:
@wferi: Yup, agree, there is no good way. Both paths are explored already (crosscompilation see openwrt,  ..., circular dependency see binutils/gcc/make/....). Removal of the map usage is probably the way, but I'm unsure how much work it means - for list we could use sys/queue.h, which is everywhere and maintained by libc maintainers.

fabbione:
@wferi I honestly have a hard time worrying too much about cross compilation. How many distros are actually cross building the HA stack? if so why? if an arch cannot build the stack "quickly" I doubt it even has a use case to run the stack itself...

wferi:
I don't think it's widespread practice, but when we broke cross-compilation in Kronosnet, I got a bug report about it. (I've got another one about Pacemaker, only half of which is fixed, because `help2man` is inherently incompatible with the concept.) People have strange ideas...

chrissie-c:
Hang on though, if we're worried about cross compilation (and TBH I'm not, very) then surely the current situation is also bad for it. If we move doxygen2man into libqb and don't use it for those man pages then that should actually solve the problem.

wferi:
Cross compilation is fixed in Kronosnet, but we don't ship doxy2xml from there, only use it internally. Not using doxygen2xml in libqb is what I referred to as the "easiest solution" above, and yes, that would provide doxygen2xml packages without any complication (from libqb or from a separate project depending on it).

fabbione:
@wferi it should be easy enough to have 2 doxygen2xml targets in the Makefile, one for local use (as we do for knet, built with the local compiler, say: doxygen2-xml-internal) and one for general shipping (cross compiled normally as any other binary). It´s a matter of enabling it in the Makefile and we should all be happy.

wferi:
Absolutely. I just meant to point out that this part is not necessary until libqb starts to use doxygen2man (see, I get the name right occasionally) so it's rightfully missing from this PR, but it's coming. On the other hand, using a separate project does not require double compilation (and also decouples the release schedules), but creates a dependency cycle when libqb starts to use doxygen2man, as Honza pointed out. In a different arrangement: moving doxygen2man into
* libqb
  * pro: less administration (no project creation)
  * con: requires double compilation pixie dust once libqb starts to use doxygen2man
* separate project:
  * pro: independent release schedules
  * con: dependency cycle once libqb starts to use doxygen2man (unless the hash implementation gets replaced)

... as I see it at the moment (feel free to correct). The choice is yours. :)

fabbione:
I vote to keep doxygen2man inside libqb and have double compilation. It´s such a small binary that 2 compilation is much much less overhead than a separate project. Clearly this require fixing this PR to handle what we already discussed.

fabbione:
All, I am moving this PR here: https://github.com/ClusterLabs/libqb/pull/388

PR #388, opened 2020-03-13
Doxygen2man

fabbione:


fabbione:
@wferi I have been talking with @chrissie-c on IRC and here is the summary:

the current build system allows to do normal and cross-compile build as-is.

Once the man pages for libqb will be auto-generated by doxygen2man, people doing normal builds will not notice any difference.

People performing cross-compilation have 2 options:
- do a normal build and install on the system and then do the cross-build to get the man pages generated automatically
- do a cross build and man pages are not generated.

This way we limit the 2 phase bootstrap of libqb only for people doing cross-compilation, that is a minority of the use base, without creating extreme complexity for people doing normal builds.

The infra is in place, just need to extend as we kill hardcoded man pages.

BtbN:
Should this really be installed along with the library though? Seems a bit out of place for libqb to install a doxygen-related tool.

kgaillot:
> Should this really be installed along with the library though? Seems a bit out of place for libqb to install a doxygen-related tool.

See the discussion on #387 which led to this one

wenningerk:
And that it is part of the libqb-repo doesn't necessarily mean you would have to package it together with the library so that it would be installed on every production cluster.

chrissie-c:
In Fedora, it has its own package.

PR #393, opened 2020-04-24
Fix error in CI tests that seems to have suddenly happened

chrissie-c:
Fix make distcheck errors in the CI.


fabbione:
ack from me

PR #397, opened 2020-05-15
test: Add unit test for ipcs_connection_auth_set()

chrissie-c:
Only if we are root

### DIFF
--- a/tests/check_ipc.c
+++ b/tests/check_ipc.c
@@ -190,6 +190,8 @@ static GSourceFuncs gio_source_funcs = {
  * 7) service availability
  *
  * 8) multiple services
+ *
+ * 9) setting perms on the sockets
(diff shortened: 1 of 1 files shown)


chrissie-c:
retest this please

chrissie-c:
This is working on all platforms now. The coverity error is just covscan not understanding that fail_if() will abort the routine if the condition fails

chrissie-c:
Note on the reproducibility of this. Currently I can only reproduce it on FreeBSD-devel by running the log.test repeatedly. Roughly 1 in 100 (but it can take longer) will fail in test_log_threads. Two threads both in qb_array_index() but at the different places in qb_log_dcs_get().

chrissie-c:
Thanks for the review.

PR #399, opened 2020-05-20
Some bugs spotted by coverity

chrissie-c:


lib/ringbuffer.c
```diff
@@ -266,7 +266,7 @@ qb_rb_open_2(const char *name, size_t size, uint32_t flags,
 	if (fd_hdr >= 0) {
 		close(fd_hdr);
 	}
-	if (rb && (flags & QB_RB_FLAG_CREATE)) {
+	if (rb && (rb->shared_hdr != MAP_FAILED) && (flags & QB_RB_FLAG_CREATE)) {
```
kgaillot, line 269: The test after this implies rb->shared_hdr could also be NULL, though I'm not sure that's true. You could either add a NULL test here or nest this inside the next test.

chrissie-c:
Looking closely into this, that NULL test is a complete red herring. Even covscan dpoesn't complain if I remove the existing one. The value of rb->shared_hdr has to be either valid or MAP_FAILED

kgaillot:
Makes sense to me, looks good

PR #400, opened 2020-06-05
array: More locking fixes

chrissie-c:
helgrind threw out a couple more locking errors in the logging/array
code.


jfriesse:
Looks good but I have one question. Is it safe to access `a->max_elements` (`qb_array_index` function) without locking?

chrissie-c:
Hmm good question. Helgrind didn't complain about it but that's not a guarantee

chrissie-c:
thanks for the review.

PR #401, opened 2020-06-10
Add qb_ipcs_alias_create() function

chrissie-c:
This creates a new IPC server destination that is simply an alias for an existing one. The idea (from Ken) is to help provide easy backwards compatibility in a lightweight way.

Posting for discussion at the moment

### DIFF
--- a/lib/ipcs.c
+++ b/lib/ipcs.c
@@ -78,6 +78,51 @@ qb_ipcs_create(const char *name,
 	return s;
 }
 
+qb_ipcs_service_t *
+qb_ipcs_alias_create(qb_ipcs_service_t *service,
+		     const char *name,
+		     enum qb_ipc_type type)
+{
+	struct qb_ipcs_service *s;
+
+	s = calloc(1, sizeof(struct qb_ipcs_service));
+	if (s == NULL) {
+		return NULL;
+	}
+	if (type == QB_IPC_NATIVE) {
+#ifdef DISABLE_IPC_SHM
+		s->type = QB_IPC_SOCKET;
+#else
+		s->type = QB_IPC_SHM;
+#endif /* DISABLE_IPC_SHM */
+	} else {
+		s->type = type;
+	}
+
+	s->pid = getpid();
+	s->needs_sock_for_poll = QB_FALSE;
+	s->poll_priority = QB_LOOP_MED;
+
+	/* Initial alloc ref */
+	qb_ipcs_ref(s);
+
+	s->service_id = service->service_id;
+	(void)strlcpy(s->name, name, NAME_MAX);
+
+	s->serv_fns.connection_accept = service->serv_fns.connection_accept;
+	s->serv_fns.connection_created = service->serv_fns.connection_created;
+	s->serv_fns.msg_process = service->serv_fns.msg_process;
+	s->serv_fns.connection_closed =	service->serv_fns.connection_closed;
+	s->serv_fns.connection_destroyed = s->serv_fns.connection_destroyed;
+	s->alias_for = service;
+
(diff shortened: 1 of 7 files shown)


lib/ipcs.c
```diff
@@ -78,6 +78,51 @@ qb_ipcs_create(const char *name,
...
+	qb_ipcs_ref(s);
+
+	s->service_id = service->service_id;
+	(void)strlcpy(s->name, name, NAME_MAX);
+
+	s->serv_fns.connection_accept = service->serv_fns.connection_accept;
+	s->serv_fns.connection_created = service->serv_fns.connection_created;
+	s->serv_fns.msg_process = service->serv_fns.msg_process;
+	s->serv_fns.connection_closed =	service->serv_fns.connection_closed;
+	s->serv_fns.connection_destroyed = s->serv_fns.connection_destroyed;
```
kgaillot, line 116: second s -> service

lib/ipcs.c
```diff
@@ -78,6 +78,51 @@ qb_ipcs_create(const char *name,
...
+
+	s->pid = getpid();
+	s->needs_sock_for_poll = QB_FALSE;
+	s->poll_priority = QB_LOOP_MED;
+
+	/* Initial alloc ref */
+	qb_ipcs_ref(s);
+
+	s->service_id = service->service_id;
+	(void)strlcpy(s->name, name, NAME_MAX);
```
kgaillot, line 110: To avoid duplication I'm thinking let qb_ipcs_create() take a NULL handlers, so this would just do that and set the handlers as below. Either that or functionize the common parts and have them both call that

chrissie-c:
Thanks. I was mostly posting it as a proof-of-concept to be sure it's actually what you wanted. It 'works' but needs a fair amount more work as you correctly point out.

chrissie-c:
Actually looking more carefully at this, there's no shared state in the alias connection structure the points to the original connection once the poll_fns have been copied. I ran valgrind over ipcserver (with qb_ipcs_destroy() calls added in the 'wrong' order) and it all seems clean. I think I'll NULL alias_for once its been used though, just to stop any use-after-free accidents happening in later PRs.

Obviously the 'connection_destroyed' typo (copy-o?) needs fixing and the duplicate code could do with tidying, but that looks like all that needs doing. Unless I'm missing something.

kgaillot:
That sounds good.

It's contrived, but the situation I think would be an issue is something like

    qb_ipcs_service_t *s1 = qb_ipcs_create(...);
    qb_ipcs_service_t *s2 = qb_ipcs_alias_create(s1, ...);
    ...
    qb_ipcs_destroy(s1);
    ...
    qb_ipcs_poll_handlers_set(s2, ...);

That seems unlikely in practice, so maybe documenting that limitation would be sufficient.

kgaillot:
> qb_ipcs_service_t *s1 = qb_ipcs_create(...);
> qb_ipcs_service_t *s2 = qb_ipcs_alias_create(s1, ...);
> ...
> qb_ipcs_destroy(s1);
> ...
> qb_ipcs_poll_handlers_set(s2, ...);

Whoops, I read the diff wrong ... I think that last line would be qb_ipcs_run(s2)

chrissie-c:
This PR got closed in the GH/CI reorganisation, but thinking about this after so much time, I'm not sure what it achieves. Connections themselves are very light inside libqb - the memory-heavy bit is the SHM part, which was being duplicated anyway. Simply creating a new service would achieve almost exactly the same thing but without adding a confusing API call that promises more than it really delivers.

kgaillot:
I never thought of that, thanks for the tip

PR #402, opened 2020-06-18
doxygen2man - Lots of new features & fixes for parsing libqb manpages

chrissie-c:
Also don't print (null) if there is no @brief description
for a function.

This means we're getting nearer to using this for libqb

### DIFF
--- a/doxygen2man/doxygen2man.1
+++ b/doxygen2man/doxygen2man.1
@@ -36,8 +36,8 @@ Write man page files to <output dir>
 .B -P
 Print PARAMS section
 .TP
-.B -s <s>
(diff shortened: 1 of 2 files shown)


chrissie-c:
Holding this for the moment, running it on libqb has exposed a few more small bugs

chrissie-c:
I got carried away and added quite a few new features. But we needed them if it's to be used for libqb.

chrissie-c:
OK, I'm going to stop fiddling with this and let someone review it now.

chrissie-c:
I have verified that the changes made here don't alter the generation of existing knet man pages at all.

PR #403, opened 2020-06-19
unix(qb_sys_circular_mmap): return -1 when mmap() returns a new addr

wladmis:
mmap() succeed when it returns a new address different from original
address. In that case errno remains unchanged and can be zero and,
therefore, cannot be used as a return value, because in this case this
leads to a memory access violation.

Found and reported by Mikhail Kulagin <m.kulagin at postgrespro dot ru>

### DIFF
--- a/lib/unix.c
+++ b/lib/unix.c
@@ -196,6 +196,9 @@ qb_sys_circular_mmap(int32_t fd, void **buf, size_t bytes)
 #endif
 
 	if (addr != addr_orig) {
+		/* explicitly set errno to non-zero because last mmap()
+		   could succeed and errno can be zero */
+		errno = EINVAL;
 		res = -errno;
 		goto cleanup_fail;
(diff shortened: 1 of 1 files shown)


chrissie-c:
retest this please

chrissie-c:
I wonder if it's worth setting errno to something useful here. Otherwise we run the risk of return -1 from the call with errno still at 0 and generating "Error: Success" type messages.

chrissie-c:
retest this please

chrissie-c:
Thanks. I'll away until Thursday, but if the CI passes I'll pull it then.

wladmis:
Please don't merge this PR for now. I see that second mmap() uses MAP_FIXED flag that should prevent this situation, but it was reported that somehow it happened that addr != addr_orig and errno == 0.

I'll check what actually happened.

PR #406, opened 2020-06-26
CI: Remove .travis.yml

chrissie-c:
Travis is too old to be any use. It's just creating work for no good
reason.

The webhook is already gone.


chrissie-c:
ACKed by fabio on IRC

PR #409, opened 2020-07-03
unix.c: use posix_fallocate()

wladmis:
Using of posix_fallocate() guarantees that, if it succeed, the
attempting to write to allocated space range does not fail because of
lack of storage space. This prevents SIGBUS when trying to write to
mmaped file and no space left.

Co-Authored-by: Ivan Zakharyaschev <imz@altlinux.org>
Reported-by: Mikhail Kulagin <m.kulagin at postgrespro dot ru>

### DIFF
--- a/lib/unix.c
+++ b/lib/unix.c
@@ -26,6 +26,9 @@
 #ifdef HAVE_SYS_MMAN_H
 #include <sys/mman.h>
 #endif
+#if defined(HAVE_FCNTL_H) && defined(HAVE_POSIX_FALLOCATE)
+#include <fcntl.h>
+#endif
 
 #include "util_int.h"
 #include <qb/qbdefs.h>
@@ -112,6 +115,14 @@ qb_sys_mmap_file_open(char *path, const char *file, size_t bytes,
 		goto unlink_exit;
 	}
 
+#ifdef HAVE_POSIX_FALLOCATE
+	if ((res = posix_fallocate(fd, 0, bytes)) != 0) {
+		errno = res;
+		res = -1 * res;
+		qb_util_perror(LOG_ERR, "couldn't allocate file %s", path);
+		goto unlink_exit;
+	}
+#else
 	if (file_flags & O_CREAT) {
 		long page_size = sysconf(_SC_PAGESIZE);
 		long write_size = QB_MIN(page_size, bytes);
@@ -138,6 +149,7 @@ retry_write:
 		}
 		free(buffer);
 	}
+#endif /* HAVE_POSIX_FALLOCATE */
 
 	return fd;
 
--- a/configure.ac
+++ b/configure.ac
@@ -259,7 +259,7 @@ AC_FUNC_CHOWN
 AC_FUNC_FORK
 AC_FUNC_MMAP
 AC_FUNC_STRERROR_R
-AC_CHECK_FUNCS([alarm fsync fdatasync ftruncate \
+AC_CHECK_FUNCS([alarm fsync fdatasync ftruncate posix_fallocate \
 		gettimeofday localtime localtime_r \
 		memset munmap socket \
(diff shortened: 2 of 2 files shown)


chrissie-c:
test this please

lib/unix.c
```diff
@@ -138,7 +151,7 @@ qb_sys_mmap_file_open(char *path, const char *file, size_t bytes,
 		}
 		free(buffer);
 	}
-
+#endif
```
imz, line 154: As discovered by @wladmis , a newline was lost unfortunately.

```suggestion
#endif

```

chrissie-c:
retest this please

chrissie-c:
I'm not sure how much optimisation that code needs (though I am in favour of posix_fallocate() part, thank you) as it's only called at rb_open() time.

imz:
> I'm not sure how much optimisation that code needs (though I am in favour of posix_fallocate() part, thank you) as it's only called at rb_open() time.

Ok, so we shouldn't bother with optimizing `memset` as something interesting.

Continuing to think about this: As far as I can see now, even doing `write` after `posix_fallocate` shouldn't be needed to get the newly allocated blocks be read as zeros. Several sources coincide on saying they will be read as zeros:

* some tests recently run by @wladmis 
* an argument [given at stackoverflow](https://stackoverflow.com/a/23508718/94687):

> The original question says something about writing zeros to the file. None of these calls write anything but metadata. If you read from space that's been preallocated but not yet written, you'll get zeros (not whatever was in that disk space previously, that would be a big security hole). You can only read up to the end of a file (the length, set by fallocate, ftruncate, or various other ways)

* [NetBSD's manpage](https://netbsd.gw.com/cgi-bin/man-cgi?posix_fallocate+2+NetBSD-7.0):

> When posix_fallocate() is applied to an unallocated region in a regular file (a ``hole''), the hole is filled and the visible contents are unaffected; both holes and newly allocated regions read as all zeros.

The old existing blocks in the file remain unzeroed, though. If we consider `qb_sys_mmap_file_open` in isolation, can this mean that this patch changes its intended behavior in non-equivalent manner?.. (I.e., `posix_fallocate` not zeroing the initial bytes.) No, because it is used from `qb_rb_open` in rungbuffer.c; there the flags are set simultanously:

```
	if (flags & QB_RB_FLAG_CREATE) {
		file_flags |= O_CREAT | O_TRUNC;
	}
```

So, the zeroing used to happen when `O_CREAT` is set, and `O_TRUNC` is set then always, too. So, first, the file is truncated to zero length, and only then extended by the new blocks.

imz:
> The old existing blocks in the file remain unzeroed, though. If we consider `qb_sys_mmap_file_open` in isolation, can this mean that this patch changes its intended behavior in non-equivalent manner?.. (I.e., `posix_fallocate` not zeroing the initial bytes.) No, because it is used from `qb_rb_open` in rungbuffer.c; there the flags are set simultanously:
> 
> ```
> 	if (flags & QB_RB_FLAG_CREATE) {
> 		file_flags |= O_CREAT | O_TRUNC;
> 	}
> ```
> 
> So, the zeroing used to happen when `O_CREAT` is set, and `O_TRUNC` is set then always, too. So, first, the file is truncated to zero length, and only then extended by the new blocks.

The only other uses of this function are from ipc_us.c, and there, too, `O_CREAT` (when zeroing happens) and `O_TRUNC` got always together:

```
	fd_hdr = qb_sys_mmap_file_open(path, r->request,
				       SHM_CONTROL_SIZE, O_RDWR);
```

or

```
	fd_hdr = qb_sys_mmap_file_open(path, r->request,
				       SHM_CONTROL_SIZE,
				       O_CREAT | O_TRUNC | O_RDWR);
```

chrissie-c:
retest this please

chrissie-c:
Merged. Thanks very much!

PR #411, opened 2020-07-16
log: add qb_log_open2() API call

chrissie-c:
Log files are created using the default umask which is not always what's wanted.

It would be neater to have this as a qb_log_ctl2() option but that call needs a handle which we can only get from qb_log_open(), so we need a new API call.

### DIFF
--- a/include/qb/qblog.h
+++ b/include/qb/qblog.h
@@ -652,6 +652,19 @@ void qb_log_format_set(int32_t t, const char* format);
 int32_t qb_log_file_open(const char *filename);
 
 /**
+ * Open a log file with set permissions
+ *
+ * These permissions are stored with the target and
+ * will also be applied to qb_log_reopen()
+ *
+ * @retval -errno on error
+ * @retval value in inclusive range QB_LOG_TARGET_DYNAMIC_START
+ *         to QB_LOG_TARGET_DYNAMIC_END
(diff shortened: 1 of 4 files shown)


wferi:
Hi Chrissie,
What do you mean by "using the default umask"? It doesn't look like `qb_log_open()` ignores the umask of its caller. 
If we absolutely can't afford to leave mode control to the user umask setting (like it's now), I'd still prefer to avoid overriding this process-wide attribute even temporarily, by using `open()` and `fdopen()` instead. The code won't be any longer and it wouldn't affect other threads, while still considering the user umask setting. If you explicitly mean to ignore the umask setting of the user (which I'd find surprising), use `chmod()` afterwards.

chrissie-c:
By 'default umask' I did mean that of the caller. If it's regarded as a bad move then I'm happy to withdraw this PR.

kgaillot:
I'm neutral on whether to do it here or via a caller's umask, but I do see one difference: the caller might change their umask after the log file is opened, and with this approach the originally requested mode will automatically be used for a later reload, but if the caller does it they will have to remember to set their desired umask before and after the reload.

chrissie-c:
Closing this as it's not needed for pacemaker and if no-one else needs it then it's just adding complexity.

PR #412, opened 2020-07-29
tests: Remove deprecated check macros

chrissie-c:
fail_if() and fail_unless() macros are deprecated in more
recent versions of check and are buggy in 0.15, so replace
them with ck_assert() instead.


chrissie-c:
Thanks.

PR #415, opened 2020-09-07
doxygen2man: Add option to read copyright line from the header file

chrissie-c:
This should help make builds reproducible.

I tried various methods of getting the date, using 'git' is no use as
it could be run from a tarball, using the file date doesn't work either
so this seems a reasonable compromise.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -123,6 +123,7 @@ BUILT_SOURCES	= .version
 
 # untaint configure.ac when modified upon obtaining the "yanked" snapshot form
 dist-hook: gen-ChangeLog
+	echo $(SOURCE_EPOCH) > $(distdir)/source_epoch
 	echo $(VERSION) | tee $(distdir)/.tarball-version | grep -Eqv '\-yank' \
 	  || sed "s/\(.*git-version-gen[^']*[']\)[^']*/\1\$$Format:%h??%D\$$/" \
 	     $(distdir)/configure.ac > $(builddir)/configure.ac-t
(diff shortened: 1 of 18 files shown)


bmwiedemann:
This change fixed the copyright and then I noticed another embedded date issue:

```diff
@@ -1,5 +1,5 @@
 .\"  Automatically generated man page, do not edit
-.TH QBUTIL.H 3 2020-09-07 "LIBQB" "libqb Programmer's Manual"
+.TH QBUTIL.H 3 2035-10-11 "LIBQB" "libqb Programmer's Manual"
 .SH NAME
```

fabbione:
I like the approach, but then please use update-copyright.sh to update all files in the git tree.

fabbione:
Also, I assume that we will need to change knet to use this new feature as well?

bmwiedemann:
One option is to use [$SOURCE_DATE_EPOCH](https://reproducible-builds.org/specs/source-date-epoch/) - there are [code snippets](https://reproducible-builds.org/docs/source-date-epoch/#c) available, too.

chrissie-c:
@bmwiedemann OK this addition patch seems to work for me, any chance you could test it on your system please?

If it works for you, do you need a new release or are the patches enough for the moment?

bmwiedemann:
commit 4ea13bb603d8c2cc40ff813b371d00d5aff795aa made it reproducible in our openSUSE environment where we have `$SOURCE_DATE_EPOCH` set, but the git codepath looked a bit wrong to me with `%at` instead of `%ct` and with the extra `$Format:` in there.

bmwiedemann:
> do you need a new release or are the patches enough for the moment

patches are enough as reproducible-builds is a long-term project and I expect there will be a new release in the next months.

chrissie-c:
I just copied the code from knet ;-)

PR #416, opened 2020-09-14
man: Tidy man pages

chrissie-c:
Fix doxygen2man to add a RETURN section even if the file has
no @return stanza.
Add support for @c tags

Quote the % signs in qb_log_format_add manpage


jfriesse (approved):
ACK by me if CI finishes successfully

PR #417, opened 2020-09-18
doxygen2man: Add support for @code blocks

chrissie-c:
Add support for @code blocks - this makes the libqb 'general' man pages look more complete now!

Because this made the resulting strings get HUGE, I eventually implemented a very simple string library to handle the mallocs so I'm not constantly mallocing huge areas or having huge amounts of data on the stack (bear in mind that get_codeline() is called recursively)


chrissie-c:
retest this please

PR #418, opened 2020-09-24
ipc: add qb_ipcc_auth_get() API call

chrissie-c:
We can't use SO_PEERCRED on the client fd when using socket IPC
becayse it's a DGRAM socket (pacemaker tries this). So provide
an API to get the server credentials that libqb has already
squirreled away for its own purposes.

Also, fix some unused-variable compiler warnings in unix.c
when building on systems withpout posix_fallocate().

### DIFF
--- a/lib/ipc_int.h
+++ b/lib/ipc_int.h
@@ -92,7 +92,6 @@ struct qb_ipcc_connection {
 	char name[NAME_MAX];
 	int32_t needs_sock_for_poll;
 	gid_t egid;
-	uid_t euid;
 	pid_t server_pid;
 	struct qb_ipc_one_way setup;
 	struct qb_ipc_one_way request;
@@ -103,6 +102,7 @@ struct qb_ipcc_connection {
 	uint32_t fc_enable_max;
 	int32_t is_connected;
 	void * context;
+	uid_t euid;
 };
 
 int32_t qb_ipcc_us_setup_connect(struct qb_ipcc_connection *c,
--- a/lib/ipc_int.h
+++ b/lib/ipc_int.h
@@ -92,6 +92,7 @@ struct qb_ipcc_connection {
 	char name[NAME_MAX];
 	int32_t needs_sock_for_poll;
 	gid_t egid;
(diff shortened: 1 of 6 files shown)


lib/ipc_int.h
```diff
@@ -92,6 +92,7 @@ struct qb_ipcc_connection {
 	char name[NAME_MAX];
 	int32_t needs_sock_for_poll;
 	gid_t egid;
+	uid_t euid;
```
kgaillot, line 95: I'd add this to the end of the struct so you don't have to bump the shared library version. Uglier, but library bumps are a hassle

lib/ipcc.c
```diff
@@ -366,6 +366,18 @@ qb_ipcc_fd_get(struct qb_ipcc_connection * c, int32_t * fd)
...
 
+int32_t
+qb_ipcc_auth_get(struct qb_ipcc_connection * c, pid_t *pid, uid_t *uid, gid_t *gid)
+{
+	if (c == NULL || pid == NULL || uid == NULL || gid == NULL) {
+		return -EINVAL;
+	}
+	*pid = c->server_pid;
+	*uid = c->euid;
+	*gid = c->egid;
```
kgaillot, line 377: For flexibility, I wouldn't require these to be non-NULL but wrap the assignments in if != NULL. That way callers can get just one or two values if that's all they need.

kgaillot (changes requested):
Nice, I think any IPC users will eventually find this handy :)

covscan is complaining but that's probably because Fabio upgraded the covscan version rather than related issues. I can't wait to see what it finds in Pacemaker ;)

chrissie-c:
Update commit in the CI now. It's still failing covscan but those are stupid errors in the test suite so I'm happy to ignore them (they'll go in the  ignore list for next time once it commited).

I have the pacemaker patch to use this APi largely ready to go too.

kgaillot:
looks good to me

kgaillot (approved):

chrissie-c:
Thanks for the review!

PR #420, opened 2020-09-24
doxygen2man: Remove horrible hack

chrissie-c:
It caused \f macros to get truncated and doesn't actually affect
the groffed output anyway


kgaillot:
looks good to me

PR #421, opened 2020-09-24
doc related fixups

wferi:
The changes in `docs/man.dox.in` silenced some "file not found" warnings, but I didn't check the actual output (the resulting `qbipcc.h.3` has no real content with 2.0.1, but maybe I looked in the wrong way).
The removal of the `http` directory is entirely untested for now, but wanted to share the idea anyway.

### DIFF
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -1,4 +1,3 @@
 man3
-html
 xml
 *.dox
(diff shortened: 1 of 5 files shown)


chrissie-c:
Thanks, that all makes sense to me. The examples don't currently get embedded in the man pages (they're a bit big I think) but it might be worth referencing them, if I can work out a way to do it sensibly.

Issue #422, opened 2020-10-02
Dont release element of qb_ipc_services when destroying ipc server


minhbq:
I encountered a crash error when I try open and close many ipcserver.

I inspected and I think the reason libqb miss release element qb_list_head from qb_ipc_services list.

Here are ipcs.c in row 213:

qb_ipcs_unref(struct qb_ipcs_service *s)
{
	int32_t free_it;

	assert(s->ref_count > 0);
	free_it = qb_atomic_int_dec_and_test(&s->ref_count);
	if (free_it) {
		qb_util_log(LOG_DEBUG, "%s() - destroying", __func__);
		free(s);
	}
}

I added more qb_list_del(&s->list) before free(s). And it works

chrissie-c:
Thanks for this, now fixed

PR #423, opened 2020-10-02
ipcs: Add missing qb_list_del when freeing server

chrissie-c:
Fix from minhbq https://github.com/ClusterLabs/libqb/issues/422


chrissie-c:
On further investigation it seems that this list isn't used anywhere, so it might as well be removed.

kgaillot:
Getting rid of qb_ipc_services looks good to me. It looks like b07b65df back in 2010 replaced the need for it with s->list but didn't drop it.

PR #424, opened 2020-10-02
ipcs: ftruncate is not support on WIN32

chrissie-c:
https://github.com/microsoft/WSL/issues/902


kgaillot:
Looks good to me (though I can't test it). It might be worthwhile to define a constant for it like HAVE_FTRUNCATE in case it comes up in some other context, but that seems unlikely.

PR #425, opened 2020-10-14
doxygen2man: Fix a couple of covscan-detected errors

chrissie-c:


doxygen2man/doxygen2man.c
```diff
@@ -498,10 +498,12 @@ static char *allcaps(const char *name)
 	static char buffer[4096] = {'\0'};
 	size_t i;
 
-	for (i=0; i< strlen(name); i++) {
-		buffer[i] = toupper(name[i]);
+	if (name) {
+		for (i=0; i< strlen(name); i++) {
+			buffer[i] = toupper(name[i]);
+		}
+		buffer[strlen(name)] = '\0';
```
kgaillot, line 505: theoretically strlen(name) could be > 4095, maybe use a variable for max(strlen(name), 4095) instead of the two strlen()'s

wferi (commented):
Wouldn't `strnlen()` be simpler? Or isn't it portable enough?

chrissie-c:
I have to admit, I'd not even heard of strnlen() thanks!

doxygen2man/doxygen2man.c
```diff
@@ -498,10 +500,13 @@ static char *allcaps(const char *name)
...
 	size_t i;
 
-	for (i=0; i< strlen(name); i++) {
-		buffer[i] = toupper(name[i]);
+	if (name) {
+		size_t len = strnlen(name, 4096);
+		for (i=0; i< len; i++) {
+			buffer[i] = toupper(name[i]);
+		}
+		buffer[len] = '\0';
```
wferi, line 508: BTW you can also shave off the `len` declaration by using `buffer[i] = 0` here (if this is your thing). :)

chrissie-c:
Fair point, but I think using len is clearer :)

kgaillot:
looks good to me

Issue #428, opened 2020-12-25
Insufficient escaping in generated man page


wferi:
The `...` in the `qbarray.h` example gets into the v2.0.2 man page unescaped, leading to the
```
$ man --warnings ./qbarray.h.3.gz >/dev/null
troff: <standard input>:13: warning: macro '..' not defined
```
and it's indeed missing from the output. In the v2.0.1 man page it's protected as `\&.\&.\&.` (which is arguably too much caution) and appears correctly.

chrissie-c:
Are you sure that 2.0.1 quoted the dots like that? docygen2man in 2.0.1 didn't support @code blocks so that part of the doc wouldn't have been in the man page anyway (and isn't on my checkout of the v2.0.1 tag).

Anyway, simplest solution is to add a format before the ellipsis so that nroff doesn't look for a macro 
https://github.com/ClusterLabs/libqb/pull/431

wferi:
Sorry, I mixed up the version numbers. It's 1.0.5 which employs `\&.\&.\&.` in `qbarray.h.3.gz`.

chrissie-c:
ahh that makes a lot more sense :)

chrissie-c:
Fixed in 47ed440397dee9133c31c4d50f7740ab59c8af63

PR #431, opened 2021-01-04
doxygen2man: fix printing of lines starting with '.'

chrissie-c:
if a line starts with a '.' (eg the '...' in qbarray.h) then
nroff thinks it's looking for a macro called '..'.

The easiest solution is to add a dummy format at the start of the line
(just adding \ seems not to work).

https://github.com/ClusterLabs/libqb/issues/428

### DIFF
--- a/doxygen2man/doxygen2man.c
+++ b/doxygen2man/doxygen2man.c
@@ -222,7 +222,14 @@ static cstring_t get_codeline(xmlNode *this_tag)
 			buffer = cstring_append_chars(buffer, " ");
 		}
 		if (strcmp((char*)sub_tag->name, "text") == 0) {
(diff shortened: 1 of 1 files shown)


wferi:
`\fN` works for me, but I couldn't find it documented anywhere. Is it really safe/portable? The usual practice of escaping the leading dot seems to be [replacing it by `\[char46]`](https://stackoverflow.com/questions/11469341/escaping-dots-in-groff).

chrissie-c:
I didn't know about that escape system, thanks. I've force-pushed that as a better fix. \fN is used all over the man pages so I'm pretty sure it works, but if \[char46] is usual practice then I will go with that.

chrissie-c:
retest this please

(CI testing)

Issue #429, opened 2020-12-25
Internal strlcpy underflows on zero buffer size


wferi:
The `QB_MIN(maxlen-1, srclen)` expression underflows for `maxlen=0` because `maxlen` is unsigned, thus `strlcpy(d,s,0)` becomes `strcpy(d,s)` contrary to the function documentation. I don't think it affects the current code base, but probably worth fixing nevertheless to avoid future accidents.

chrissie-c:
Fixed in 5097155bdf39bd18a778b985354b0b38fd499f55

PR #432, opened 2021-01-04
strlcpy: Check for maxlen underflow

chrissie-c:
https://github.com/ClusterLabs/libqb/issues/429


chrissie-c:
Yes, that does seem to be a more sensible behaviour.(every bug in libqb seems to lead to two more :)

wferi:
Seems correct to me now. At the risk of breaking it again: do you still need the additional `len2cpy` check? Wouldn't an unconditional `memcpy(dest, src, len2cpy)` do? (Optionally with the `len2cpy` definition moved inside the conditional branch.)
Or just use libbsd on Linux. :)

chrissie-c:
It's tempting to use libbsd, certainly! But one of the aims of libqb is to be as dependency-free as possible.

PR #434, opened 2021-01-22
ipcc: Have a few goes at tidying up after a dead server

chrissie-c:
This is an attempt to make sure that /dev/shm is cleaned up when a
server exits unexpectedly. Normally it's the server's responsibility
to tidy up sockets, but if it crashes or is killed with SIGKILL then
the client (us) makes a reasonable attempt to tidy up the server sockets
we have connected. The extra delay here just gives the server chance to
disappear fully. As a client we can get here pretty quickly but shutting
down a large server may take a little longer even when SIGKILLed.
The 1/100th of a second is an arbitrary delay (of course) but seems to
catch most servers in 2 tries or less.

See https://bugzilla.redhat.com/show_bug.cgi?id=1614166 for more info.
And yes, I'm expecting this to be controversial and anyone with better ideas is welcome.

### DIFF
--- a/lib/ipc_shm.c
+++ b/lib/ipc_shm.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2010 Red Hat, Inc.
+ * Copyright (C) 2010-2021 Red Hat, Inc.
  *
  * Author: Angus Salkeld <asalkeld@redhat.com>
  *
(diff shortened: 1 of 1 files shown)


jfriesse (approved):
Seems to solve problem so ACK from me.

kgaillot:
Looks good to me too.

BTW a separate issue for libqb is that usleep() is considered obsolete (no longer in the POSIX standard, which defines nanosleep()). I don't know of any platforms where it's a problem but it might crop up one day.

wferi:
If `/dev/shm` wasn't hard-coded (and if I understand the issue right), this could be neatly solved by the `RuntimeDirectory` systemd directive used by all IPC server units. As things stand, `ExecStopPost` could do the required cleanup instead. I'm not sure it has all required information (for example server PID) available, though...

PR #435, opened 2021-01-27
tests: Fix up resources.test

chrissie-c:
resources.test has not checked the right filenames for a while.
Fix this, and also make sure we don't count (but do remove) the dlock
test files.


jfriesse (approved):
Looks good so ACK.

PR #436, opened 2021-01-27
timers: Add some locking

chrissie-c:
Fix several locking issues reported by helgrind

### DIFF
--- a/include/tlist.h
+++ b/include/tlist.h
@@ -36,6 +36,7 @@ static int64_t timerlist_hertz;
 
 struct timerlist {
 	struct qb_list_head timer_head;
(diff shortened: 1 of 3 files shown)


jfriesse (approved):
Looks great and makes libqb a bit more multi thread friendly so ACK. I would just consider increasing sleep timeout in test to prevent false failures because of clogged CI.

chrissie-c:
Thanks for the review. merged.

sdake:
While this patch is technically correct, unless the title was “all” it should be reverted. This sets a precedent this library supports concurrency. It doesn’t, and this small change doesn’t move things any closer, if concurrency supports is a goal.

PR #437, opened 2021-02-17
RFC: Add locking to map functions

chrissie-c:
In my investigations into the thread-safety of libqb it turns out
that ... it isn't.

This patch adds locking calls to all of the map primitives to make
them thread-safe (or at least pass a lot of helgrind tests).

There is a performance hit to this which varies depending on the type of
map being used, skiplists seem to be worst hit. So the discussion is
really is it worth the hit for thread-safety (which is already an
advertised feature!) or decide it's not worth it as I've only had 1 bug
report on thread-safety (which was not even in the maps, it was in
timers) in ages.

with locking:
real    6m25.355s - for 100 runs run of map.test
user    5m17.371s
sys     0m53.592s

From a single run, but reasonably representative
check_map.c:692 info     test_skiplist_load_fn   1073733.25 puts/sec (100000/0.093133s)
check_map.c:710 info     test_skiplist_load_fn    448605.53 gets/sec (100000/0.222913s)
check_map.c:722 info     test_skiplist_load_fn   6689410.50 iters/sec (100000/0.014949s)
check_map.c:741 info     test_skiplist_load_fn     38667.70 dels/sec (100000/2.586138s)
check_map.c:692 info    test_hashtable_load_fn   1938360.12 puts/sec (100000/0.051590s)
check_map.c:710 info    test_hashtable_load_fn    558843.44 gets/sec (100000/0.178941s)
check_map.c:722 info    test_hashtable_load_fn   3511976.00 iters/sec (100000/0.028474s)
check_map.c:741 info    test_hashtable_load_fn    560283.75 dels/sec (100000/0.178481s)
check_map.c:692 info         test_trie_load_fn    866708.88 puts/sec (100000/0.115379s)
check_map.c:710 info         test_trie_load_fn    575440.25 gets/sec (100000/0.173780s)
check_map.c:722 info         test_trie_load_fn   2461235.50 iters/sec (100000/0.040630s)
check_map.c:741 info         test_trie_load_fn    461812.72 dels/sec (100000/0.216538s)

without locking (current code):
real    4m42.447s  - for 100 runs of map.test
user    3m42.478s
sys     0m51.580s

From a single run, but reasonably representative
check_map.c:683 info     test_skiplist_load_fn   1838269.12 puts/sec (100000/0.054399s)
check_map.c:701 info     test_skiplist_load_fn    542690.75 gets/sec (100000/0.184267s)
check_map.c:713 info     test_skiplist_load_fn  11125945.00 iters/sec (100000/0.008988s)
check_map.c:732 info     test_skiplist_load_fn     65102.60 dels/sec (100000/1.536037s)
check_map.c:683 info    test_hashtable_load_fn   1982317.75 puts/sec (100000/0.050446s)
check_map.c:701 info    test_hashtable_load_fn    590521.00 gets/sec (100000/0.169342s)
check_map.c:713 info    test_hashtable_load_fn   9184423.00 iters/sec (100000/0.010888s)
check_map.c:732 info    test_hashtable_load_fn    640881.81 dels/sec (100000/0.156035s)
check_map.c:683 info         test_trie_load_fn   1213032.88 puts/sec (100000/0.082438s)
check_map.c:701 info         test_trie_load_fn    599635.44 gets/sec (100000/0.166768s)
check_map.c:713 info         test_trie_load_fn   3376781.25 iters/sec (100000/0.029614s)
check_map.c:732 info         test_trie_load_fn    523409.50 dels/sec (100000/0.191055s)

### DIFF
--- a/lib/hashtable.c
+++ b/lib/hashtable.c
@@ -21,6 +21,7 @@
 #include "os_base.h"
 
 #include <qb/qbmap.h>
+#include <pthread.h>
 #include "util_int.h"
 #include "map_int.h"
 #include <qb/qblist.h>
@@ -37,6 +38,7 @@ struct hash_node {
 
 struct hash_bucket {
 	struct qb_list_head list_head;
+	pthread_mutex_t bkt_lock;
 };
 
 struct hash_table {
@@ -45,6 +47,9 @@ struct hash_table {
 	uint32_t order;
 	uint32_t hash_buckets_len;
 	struct qb_list_head notifier_head;
+	/* Lock ordering:
+	   If you need a bucket lock and this lock, always get the bucket lock first */
+	pthread_mutex_t ht_lock;
 	struct hash_bucket hash_buckets[0];
 };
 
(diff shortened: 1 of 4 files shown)


jfriesse:
The performance hit seems to be pretty bad (and some numbers are interesting - like I would expect pthread to make app spent more time in sys but it's not happening). I don't think it would affect real app that much but it will need some real life testing.

Just an idea - what about having both options? So either one API for really thread safe operations and second one which is thread unsafe? Or it can be some kind of flag when map/... is created? Not sure tho if default should be thread safe (would make sense to match docs) or thread unsafe (for "compatibility")? How much bigger (messier) it would make patch?

chrissie-c:
The patch could be quite sane, just define versions of pthread_mutex_* as inlines that check a variable before doing the locking or not. The big question is how we set that in the API of course.

I suspect the lack of extra 'sys' time is down to the use of futexes that do all of the 'fast path' work in userspace without a syscall.

chrissie-c:
Hmm, the CI has thrown up a lot of extra things that helgrind didn't. This will take quite a bit more work.

chrissie-c:
Closing this as it's impact is greater than it's function. Despite mentioning thread-safety a few times in the documentation it's clear that libqb is NOT thread safe and was never written to be thread-safe. Retro-fitting it is expensive, error-prone and benefiting no users to the library that we are aware of.

PR #438, opened 2021-02-24
blackbox: Sanitize items read from the blackbox header

chrissie-c:
covscan complained we don't check the blackbox header when
reading it in - quite reasonably

(first pass for covscan approval)


kgaillot:
I'm not familiar with the blackbox code but the general idea looks good to me

fabbione (approved):

PR #440, opened 2021-03-16
build: Fix undefined pthread reference.

orbea:
When using slibtool (https://github.com/midipix-project/slibtool) instead of GNU libtool the build fails with undefined references to pthread.
```
/usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: bmcpt.o: undefined reference to symbol 'pthread_create@@GLIBC_2.2.5'
/usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: /lib64/libpthread.so.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```
This works with GNU libtool because it silently filters out `-no-undefined` while slibtool does not.

This can be easily fixed by added `$(PTHREAD_CFLAGS)` and `$(PTHREAD_LIBS)` where applicable.

Also see this downstream issue: https://bugs.gentoo.org/775605

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -38,7 +38,8 @@ bmc_SOURCES = bmc.c
 bmc_LDADD = $(top_builddir)/lib/libqb.la
 
 bmcpt_SOURCES = bmcpt.c
-bmcpt_LDADD = $(top_builddir)/lib/libqb.la
+bmcpt_CFLAGS = $(PTHREAD_CFLAGS)
+bmcpt_LDADD = $(PTHREAD_LIBS) $(top_builddir)/lib/libqb.la
 
 bms_SOURCES = bms.c
 bms_CFLAGS = $(GLIB_CFLAGS)


fabbione:
ok to test

fabbione (approved):

fabbione:
@orbea is slibtool the default on gentoo? if so it might be worth adding Gentoo to the CI nodes pool.

orbea:
> is slibtool the default on gentoo?

No, its currently being worked one for future integration. Whether it becomes the default or remains just an option in the future I am not sure. This issue tracks current slibtool issues found in gentoo:

https://bugs.gentoo.org/765709

Most, but not all issues are issues are the result of GNU libtool silently hiding bugs for so long...

fabbione:
retest this please
(due to CI network hickup)

fabbione:
@orbea thanks for the pointer. slibtools seems to be available in different distributions so perhaps we can just enable a build with it to avoid regression in future.

chrissie-c:
Looks good, thanks for that patch, and to Fabio for reviewing.

jnpkrn:
> Most, but not all issues are the result of GNU libtool silently hiding
> bugs for so long...

Ah, cannot but to relate to this, especially since downstreams tend to
add fuel to the fire with diverging = less predictable tweaks.
I remeber we have met this before, looks like it was at this occasion:
https://github.com/ClusterLabs/libqb/commit/20246f544f74866885f4d9491f74152bd95e7786#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810R624-R634

(perhaps also related: https://github.com/ClusterLabs/libqb/pull/323#issuecomment-422015795)

PR #443, opened 2021-06-30
doxygen2man: print structure descriptions

chrissie-c:


doxygen2man/cstring.c
```diff
@@ -51,6 +51,19 @@ char *cstring_to_chars(cstring_t cstring)
 	return strdup(h->the_string);
 }
 
+size_t cstring_len(cstring_t cstring)
+{
+	struct cstring_header *h = (struct cstring_header *)(char *)cstring;
```
wferi, line 56: Hi @chrissie-c, why do you use these double casts?

chrissie-c: TBH I can't remember now! I suspect it was to do with the compiler (or maybe covscan) complaining about alignment. I should check if they are still needed.

PR #444, opened 2021-08-09
Fix pthread returns

chrissie-c:
pthread calls do not set errno, they return the error directly


jfriesse (approved):
Looks sensible so ACK

chrissie-c:
Thanks for the review

PR #447, opened 2021-11-10
poll: Don't log in a signal handler

chrissie-c:
qb_log calls malloc() and probably many other non-signal-safe
functions, so don't call it in the signal handler.

Thanks to Honza for spotting this


jfriesse (approved):
ACK from me... not so happy about error not covered but there seems to be no reasonable way to report error so this is probably best we can do.

chrissie-c:
Yeah, removing error messages is not really a good thing, but if the alternative is potential hang, we just need to balance things :/

PR #454, opened 2022-01-12
Tidy some scripts

chrissie-c:
Errors reported by Centos covscan

/bin/bash in start.test is for echo -n

### DIFF
--- a/tests/resources.test
+++ b/tests/resources.test
@@ -6,8 +6,8 @@ IPC_NAME=`cat ipc-test-name 2>/dev/null`
 for d in /dev/shm /var/run $SOCKETDIR; do
 
 	# Tidy up the deadlock checker sockets first
-	dlocks=$(find $d -name qb-*-test_*dlock*${IPC_NAME}* -size +0c 2>/dev/null)
-	if [ `echo $dlocks|wc -w` -eq $(($SOCKS_PER_PROCESS * 6)) ]; then
+	dlocks=$(find $d -name "qb-*-test_*dlock*${IPC_NAME}*" -size +0c 2>/dev/null)
+	if [ "`echo $dlocks|wc -w`" -eq $(($SOCKS_PER_PROCESS * 6)) ]; then
 		rm $dlocks
 	elif [ -n "${dlocks}" ]; then
 		echo
@@ -17,14 +17,14 @@ for d in /dev/shm /var/run $SOCKETDIR; do
 	fi
 
 	# Now look for other leftovers
-	leftovers=$(find $d -name qb-*-test_*${IPC_NAME}* -size +0c 2>/dev/null | wc -l)
+	leftovers=$(find $d -name "qb-*-test_*${IPC_NAME}*" -size +0c 2>/dev/null | wc -l)
 	if [ "${leftovers}" -gt 0 ]; then
 		echo
 		echo "Error: shared memory segments not closed/unlinked"
 		echo
 		RETURN=1
 	fi
-	leftovers="$(find $d -name qb-*-test_*${IPC_NAME}* -size 0c 2>/dev/null)"
+	leftovers="$(find $d -name "qb-*-test_*${IPC_NAME}*" -size 0c 2>/dev/null)"
 	if [ "$(printf '%s\n' "${leftovers}" | wc -l)" -eq $(($SOCKS_PER_PROCESS * 2)) ]; then
 		echo
 		echo "There were some empty leftovers (expected), removing them"
--- a/tests/start.test
+++ b/tests/start.test
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 #
 # Generate a unique(ish) name for the IPCs we will use in the tests and


jfriesse (approved):
Looks good so ACK by me

wferi:
> /bin/bash in start.test is for echo -n

Just out of curiosity: wouldn't something as simple as `printf "$$-$RANDOM"` suffice?

chrissie-c:
@wferi yes, that makes a lot more sense

wenningerk:
or for the paranoid use mktemp (in a common temp-file) to assure it is really unique ...

chrissie-c:
@wenningerk libqb already does this for directories created by the client. This name is the unique name for the service (that clients connect to). These need to be unique for testing so we can run >1 instance at the same time. In normal running those names are well-known names (like cmap/votequorum).

wenningerk:
> @wenningerk libqb already does this for directories created by the client. This name is the unique name for the service (that clients connect to). These need to be unique for testing so we can run >1 instance at the same time. In normal running those names are well-known names (like cmap/votequorum).

Yep - running multiple tests (with multiple server instances) in parallel was what I was thinking of + the unlikely case that random numbers might be the same - but as said just for the paranoid ;-)

chrissie-c:
The server names are in the UNIX-domain socket namespace, not a filesystem. So mk[s]temp is no help here. There is the option for libqb to use filesystem sockets (which is only used AFAIK for inter-container communication). I have  PR somewhere for testing that ;-)

chrissie-c:
Ahhh! In that case I'll put the date thing back in again. Thanks for checking.

chrissie-c:
Having said that date+%N doesn't work on BSD anyway (it just prints 'N') so I'll do some work on that.

And this was going to be a simple PR :)

wenningerk:
Maybe the mktemp approach isn't that stupid after all - at least it exists on BSD ;-)
The nano-seconds are a special thing. They don't exist in the c-library as well.
Remember the pain that I had to implement them so that the alert-agents could
define a high-resolution-timestamp. @wferi: thanks for finding a bug with that
back then iirc.

chrissie-c:
retest this please

wferi:
> This name is the unique name for the service (that clients connect to). These need to be unique for testing so we can run >1 instance at the same time. In normal running those names are well-known names (like cmap/votequorum).

Why not use the PID of the make process, though? You could even communicate it via the environment and get rid of `ipc-test-name`, `start.test` and the randomness: `export IPC_TEST_NAME != printf 'what-%d-ever' "$$PPID"`.

wferi:
On the other hand, isn't the current `start.test` method racy under parallel builds (`make -j`)?

chrissie-c:
@wferi something racy in libqb? Surely not? ;-)

I'm not sure what it would race with though, the tests are run in series even with make -j

wferi:
> the tests are run in series even with `make -j`

Well, this is what I get in `start.log` if I append `sleep 0.1; ps xf` to the end of `start.test`:
```
  PID TTY      STAT   TIME COMMAND
31857 pts/0    S      0:00 /bin/bash
  795 pts/0    S+     0:00  \_ make -j 16 check
  813 pts/0    S+     0:00      \_ make check-TESTS
  827 pts/0    S+     0:00          \_ /bin/bash -c set +e; bases='start.log array.log map.log rb.log list.log log.log blackbox-segfault.sh.log loop.log ipc.log resources.log'; bases=`for i in $bases; do echo $i; done | sed 's/\.log$//'`; bases=`echo $bases`; \ log_list=`for i in $bases; do echo $i.log; done`; \ trs_list=`for i in $bases; do echo $i.trs; done`; \ log_list=`echo $log_list`; trs_list=`echo $trs_list`; \ make  test-suite.log TEST_LOGS="$log_list"; \ exit $?;
  836 pts/0    S+     0:00              \_ make test-suite.log TEST_LOGS=start.log array.log map.log rb.log list.log log.log blackbox-segfault.sh.log loop.log ipc.log resources.log
  845 pts/0    S+     0:00                  \_ /bin/bash ../build-aux/test-driver --test-name start.test --log-file start.log --trs-file start.trs --color-tests yes --enable-hard-errors yes --expect-failure no -- ./start.test
  880 pts/0    S+     0:00                  |   \_ /bin/sh ./start.test
 1121 pts/0    R+     0:00                  |       \_ ps xf
  859 pts/0    S+     0:00                  \_ /bin/bash ../build-aux/test-driver --test-name log.test --log-file log.log --trs-file log.trs --color-tests yes --enable-hard-errors yes --expect-failure no -- ./log.test
  890 pts/0    S+     0:00                  |   \_ /build/libqb-FX3vsk/libqb-2.0.4/tests/.libs/log.test
 1023 pts/0    Sl     0:00                  |       \_ /build/libqb-FX3vsk/libqb-2.0.4/tests/.libs/log.test
  860 pts/0    S+     0:00                  \_ /bin/bash ../build-aux/test-driver --test-name blackbox-segfault.sh --log-file blackbox-segfault.sh.log --trs-file blackbox-segfault.sh.trs --color-tests yes --enable-hard-errors yes --expect-failure no -- ./blackbox-segfault.sh
  897 pts/0    S+     0:00                  |   \_ /bin/sh ./blackbox-segfault.sh
  862 pts/0    S+     0:00                  \_ /bin/bash ../build-aux/test-driver --test-name loop.test --log-file loop.log --trs-file loop.trs --color-tests yes --enable-hard-errors yes --expect-failure no -- ./loop.test
  893 pts/0    S+     0:00                  |   \_ /build/libqb-FX3vsk/libqb-2.0.4/tests/.libs/loop.test
 1085 pts/0    S      0:00                  |       \_ /build/libqb-FX3vsk/libqb-2.0.4/tests/.libs/loop.test
  863 pts/0    S+     0:00                  \_ /bin/bash ../build-aux/test-driver --test-name ipc.test --log-file ipc.log --trs-file ipc.trs --color-tests yes --enable-hard-errors yes --expect-failure no -- ./ipc.test
  891 pts/0    S+     0:00                      \_ /build/libqb-FX3vsk/libqb-2.0.4/tests/.libs/ipc.test
 1003 pts/0    S      0:00                          \_ /build/libqb-FX3vsk/libqb-2.0.4/tests/.libs/ipc.test
 1007 pts/0    Z      0:00                              \_ [ipc.test] <defunct>
[...]
```
Doesn't this mean that the file reads in the actual tests race with the write in `start.test`?

chrissie-c:
oh good, another bug in the test suite!

chrissie-c:
I think, for the moment, I'm going to leave that - it's outside the scope of this PR and I don't currently have the motivation to muck around with Makefile.am files (TBH I *never* have the motivation to much about with Makefile.am files)

chrissie-c:
Thanks, I will accept patches to make parallel builds work properly ;-)

wferi:
Speaking of parallel test, has the `ipc-test-name` to be shared among the tests? Or can each test use a different `ipc-test-name` as long as they are unique?

fabbione:
@wferi I remember looking at it at some point.. I don´t remember the details of the implementation, but I do offer one piece of advice. Some of the libqb tests are very CPU intensive and they have timeouts and alike.. running too many of them in parallel might cause too many false errors. I have experienced this in CI when running multiple libqb builds in parallel (not even the test suite), very often they would fail due to timeouts. CI now serializes libqb builds to avoid that false error.

chrissie-c:
@wferi In theory different tests could use a different ipc-test-name. In practice though, only test.ipc actually does any IPC testing so it's not really relevant :)

PR #455, opened 2022-01-12
tests: Run IPC with use-filesystem-sockets active

chrissie-c:
(re-opening PR 380 that got lost in the Great Git Transition of 2021)

Provide an LD_PRELOAD library that simulates the presence
of /etc/libqb/use-filesystem-sockets so that we can test
that functionality without actually having the file on
the system and affecting everything else running on the
box.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -176,7 +176,7 @@ PKG_CHECK_MODULES([libxml], [libxml-2.0])
 # don´t build the man pages
 if test "x$cross_compiling" = "xno"; then
   AM_CONDITIONAL([BUILD_MAN], [true])
-  DOXYGEN2MAN="\"\$(abs_builddir)/../doxygen2man/doxygen2man\""
+  DOXYGEN2MAN="\$(abs_builddir)/../doxygen2man/doxygen2man"
 else
   AC_CHECK_PROGS([DOXYGEN2MAN], [doxygen2man])
   if test "x$DOXYGEN2MAN" = "x"; then
--- a/configure.ac
+++ b/configure.ac
@@ -176,7 +176,7 @@ PKG_CHECK_MODULES([libxml], [libxml-2.0])
 # don´t build the man pages
 if test "x$cross_compiling" = "xno"; then
   AM_CONDITIONAL([BUILD_MAN], [true])
-  DOXYGEN2MAN="\$(abs_builddir)/../doxygen2man/doxygen2man"
+  DOXYGEN2MAN="\"\$(abs_builddir)/../doxygen2man/doxygen2man\""
 else
   AC_CHECK_PROGS([DOXYGEN2MAN], [doxygen2man])
   if test "x$DOXYGEN2MAN" = "x"; then
--- a/tests/ipc_sock.test
+++ b/tests/ipc_sock.test
@@ -1,0 +1,19 @@
+#!/bin/sh
+#
+# Run the IPC tests under the stat wrapper,
+# this simulates /etc/libqb/use-filesystem-sockets existing
+# so we can test both options without breaking other things
+# that might be running on this system
+#
+if [ "`uname -s`" = "Linux" ]
+then
+  if [ -f `pwd`/.libs/libstatwrapper.so ]
+  then
+    export LD_PRELOAD=`pwd`/.libs/libstat_wrapper.so
+  else
+    export LD_PRELOAD=`pwd`/libstat_wrapper.so
+  fi
+  ./ipc.test
+else
+  exit 0
+fi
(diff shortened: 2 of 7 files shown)


chrissie-c:
retest this please

chrissie-c:
retest this please

tests/libstat_wrapper.c
```diff
@@ -0,0 +1,35 @@
...
+
+int __xstat(int __ver, const char *__filename, struct stat *__stat_buf)
+{
+#if defined(QB_LINUX) || defined(QB_CYGWIN)
+	static int opened = 0;
+	static void *dlhandle;
+	static int (*real_xstat)(int __ver, const char *__filename, void *__stat_buf);
+
+	if (!opened) {
+		dlhandle = dlopen(LIBC_SO, RTLD_NOW);
```
wenningerk, line 23: Had been doing that in the past with dlopen_fatal(RTLD_NEXT, ...) at least in the assumption it would then get me the symbol coming next in the search order instead of directly jumping to libc.
And instead of keeping the handle in a static variable (when using an explicit library - with RTLD_NEXT there shouldn't be a need) I closed it after getting the symbols I needed.

tests/libstat_wrapper.c
```diff
@@ -0,0 +1,35 @@
...
+		real_xstat = dlsym(dlhandle, "__xstat");
+		opened = 1;
+	}
+
+	if (strcmp(__filename, FORCESOCKETSFILE) == 0) {
```
wenningerk, line 28: __xstat seems to return -1 with errno==EFAULT in case the filename points to invalid memory instead of segfaulting.
Don't know if we need it that generic here but we might get called by unexpected callers and get confusing results.

tests/libstat_wrapper.c
wenningerk: Sry for the above - that is of course dlsym_fatal.
And that again is just my wrapper for dlsym exiting out if it doesn't find the symbol.
The comment in parentheses already points to something being fishy here ;-)

chrissie-c:
That last force push was just a rebase to fix a covscan error in CI

tests/ipc_sock.test
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+#
+# Run the IPC tests under the stat wrapper,
+# this simulates /etc/libqb/use-filesystem-sockets existing
+# so we can test both options without breaking other things
+# that might be running on this system
+#
+if [ "`uname -s`" = "Linux" ]
```
fabbione, line 8: General comment here, also spotted below in other test scripts. It´s best to avoid the `command.. ` format and use instead "$(command..)". When using `` the command is forked in another shell that resets the envvar to default. This could cause unexpected behavior. Similar for `pwd` etc. forking also uses more resources. With $(cmd) the command is executed within the same shell.

tests/ipc_sock.test
```diff
@@ -0,0 +1,19 @@
...
+# Run the IPC tests under the stat wrapper,
+# this simulates /etc/libqb/use-filesystem-sockets existing
+# so we can test both options without breaking other things
+# that might be running on this system
+#
+if [ "$(uname -s)" = "Linux" ]
+then
+  if [ -f $(pwd)/.libs/libstat_wrapper.so ]
+  then
+    export LD_PRELOAD=$(pwd)/.libs/libstat_wrapper.so
```
fabbione, line 12: best to have:
export LD_PRELOAD="$(pwd)/.libs/libstat_wrapper.so" in case pwd contains spaces... it doesn´t happen in CI or any sane systems.. but it can happen :)

fabbione (approved):

fabbione (approved):

configure.ac
```diff
@@ -176,7 +176,7 @@ PKG_CHECK_MODULES([libxml], [libxml-2.0])
 # don´t build the man pages
 if test "x$cross_compiling" = "xno"; then
   AM_CONDITIONAL([BUILD_MAN], [true])
-  DOXYGEN2MAN="\$(abs_builddir)/../doxygen2man/doxygen2man"
+  DOXYGEN2MAN="\"\$(abs_builddir)/../doxygen2man/doxygen2man\""
```
kgaillot, line 179: For future-proofing it would be better to leave this alone and quote $(DOXYGEN2MAN) in docs/Makefile.am, otherwise if someone ever uses "$DOXYGEN2MAN" it will fail in a confusing way

kgaillot (commented):
Looks good to me too

fabbione:
retest this please

fabbione (approved):

PR #459, opened 2022-02-23
test: Clean /dev/shm a bit better

chrissie-c:
This isn't perfect, but it does tidy more of /dev/shm than
previously. Because some of the tests leave empty directories
we have no way of telling (in resources.test) whether they
belong to this test run, another test run, or a running
application.

### DIFF
--- a/tests/resources.test
+++ b/tests/resources.test
@@ -2,16 +2,22 @@
 RETURN=0
 SOCKS_PER_PROCESS=3
 
+tidy_qb_dirs()
+{
+    for dd in "$@"; do
+	rm $dd
+	rmdir `dirname $dd` 2> /dev/null
+    done
+}
+
+
 IPC_NAME=`cat ipc-test-name 2>/dev/null`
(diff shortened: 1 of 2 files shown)


tests/resources.test
```diff
@@ -28,8 +31,12 @@ for d in /dev/shm /var/run $SOCKETDIR; do
 	if [ "$(printf '%s\n' "${leftovers}" | wc -l)" -eq $(($SOCKS_PER_PROCESS * 2)) ]; then
 		echo
 		echo "There were some empty leftovers (expected), removing them"
-		echo "${leftovers}" | tee /dev/stderr | xargs rm
+		echo "${leftovers}" | tee /dev/stderr
```
kgaillot, line 34: It seems odd to output just the names to both stdout and stderr, I would expect all the echos to go to one place. echo blah >&2 goes to just stderr

kgaillot (commented):
looks good to me, though I didn't test it

fabbione (approved):

PR #460, opened 2022-03-02
ipcc: Fix errno returned from qb_ipcc_connect

chrissie-c:
The errno value from qb_ipcc_connect was incorrectly negated
when I introduced qb_ipcc_async_connect()


kgaillot (approved):
looks good to me

PR #461, opened 2022-03-14
unix: Don't fail on FreeBSD running ZFS

chrissie-c:
ZFS doesn't support posix_fallocate() so libqb IPC or RB would
always fail with EINVAL.

As there seems to be no prospect of a more useful return code,
trap it in a QB_BSD #ifdef. That way if we do have actual errors
in the posix_fallocate() call the Linux tests should still find them.

Also, stick a small sleep in the test_ipc_disconnect_after_created
test to allow the server to shutdown before killing it with SIGTERM
and causing a test failure. all the other uses of it seem to have this
sleep!

### DIFF
--- a/lib/unix.c
+++ b/lib/unix.c
@@ -132,6 +132,11 @@ qb_sys_mmap_file_open(char *path, const char *file, size_t bytes,
 		if (res == EINTR) {
 			qb_util_log(LOG_DEBUG, "got EINTR trying to allocate file %s, retrying...", path);
 			continue;
+#ifdef QB_BSD
+		} else if (res == EINVAL) { /* posix_fallocate() fails on ZFS
+					       https://lists.freebsd.org/pipermail/freebsd-current/2018-February/068448.html */
+			qb_util_log(LOG_DEBUG, "posix_fallocate returned EINVAL - running on ZFS?");
+#endif
 		} else if (res != 0) {
 			errno = res;
 			res = -1 * res;
--- a/lib/unix.c
+++ b/lib/unix.c
@@ -69,6 +69,44 @@ open_mmap_file(char *path, uint32_t file_flags)
 	return open(path, file_flags, 0600);
 }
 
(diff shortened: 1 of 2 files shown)


lib/unix.c
```diff
@@ -132,6 +132,11 @@ qb_sys_mmap_file_open(char *path, const char *file, size_t bytes,
 		if (res == EINTR) {
 			qb_util_log(LOG_DEBUG, "got EINTR trying to allocate file %s, retrying...", path);
 			continue;
+#ifdef QB_BSD
+		} else if (res == EINVAL) { /* posix_fallocate() fails on ZFS
+					       https://lists.freebsd.org/pipermail/freebsd-current/2018-February/068448.html */
+			qb_util_log(LOG_DEBUG, "posix_fallocate returned EINVAL - running on ZFS?");
+#endif
```
kgaillot, line 139: Another possibility would be to do what's currently in the #else block if posix_fallocate() fails, as a fallback (regardless of platform). Not sure if that's better though ...

tests/check_ipc.c
```diff
@@ -1959,8 +1959,8 @@ test_ipc_disconnect_after_created(void)
 		ck_assert_int_eq(res, -ENOTCONN);
 	}
 	ck_assert_int_eq(QB_FALSE, qb_ipcc_is_connected(conn));
-
 	qb_ipcc_disconnect(conn);
+	sleep(1); /* Give it time to stop */
 	kill_server(pid);
```
kgaillot, line 1964: If you want to make sure this doesn't get overlooked, you could make the sleep time an argument to kill_server()

kgaillot (commented):
Looks good

kgaillot:
looks good

chrissie-c:
Thanks Ken.

if you get time, could you also please have a look at https://github.com/ClusterLabs/libqb/pull/455 - it's been hanging around for a while and keeps getting overlooked, but I think we should be testing this feature as I believe it's till used in container (and if not, we can remove it!)

PR #466, opened 2022-06-08
tests: Make ipc test more portable

chrissie-c:
fix a couple of errors reported by covscan

1. /usr/lib64/libqb/tests/resources.test:8:34: warning[SC2039]: In POSIX sh, UID is undefined.
1. /usr/lib64/libqb/tests/resources.test:8:30: warning[SC2166]: Prefer [ p ] && [ q ] as [ p -a q ] is not well defined.
1. /usr/lib64/libqb/tests/ipc_sock.test:10:34: warning[SC2039]: In POSIX sh, UID is undefined.
1. /usr/lib64/libqb/tests/ipc_sock.test:10:30: warning[SC2166]: Prefer [ p ] && [ q ] as [ p -a q ] is not well defined.

### DIFF
--- a/tests/ipc_sock.test
+++ b/tests/ipc_sock.test
@@ -7,7 +7,7 @@
 #
 # This creates sockets in /var/run so needs to be root
 #
-if [ "$(uname -s)" = "Linux" -a "$UID" = "0" ]
+if [ "$(uname -s)" = "Linux" ] && [ "`id -u`" = "0" ]
 then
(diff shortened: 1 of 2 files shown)


fabbione:
ack

chrissie-c:
The new code covscan warnings can be waived.
1) rand() is called in a test program that's is nothing to do with crypto
2) max_msg_size can never be -1 or it's unsigned equivalent unless the user explicitly makes it so with a library call.

PR #467, opened 2022-06-15
tests: clean up the last of the empty directories

chrissie-c:
This should, FINALLY, allow the libqb tests to run
and properly clean up up after themselves.


chrissie-c:
Thanks Ken.

PR #468, opened 2022-06-16
doxygen2man: Fix function parameter alignment

chrissie-c:
Also, prevent truncation of parameter descriptions in the
body of the man page due to strtok() usage.


chrissie-c:
```
       int32_t qb_hdb_handle_create(
           struct qb_hdb   *hdb,  /* the database instance */
           int32_t          instance_size,            /* size of the object to malloc */
           qb_handle_t     *handle_id_out            /* new handle */
       );

```
becomes

```
       int32_t qb_hdb_handle_create(
           struct qb_hdb   *hdb,            /* the database instance */
           int32_t          instance_size,  /* size of the object to malloc */
           qb_handle_t     *handle_id_out   /* new handle */
       );


```

fabbione (approved):
Untested but the patch looks good

PR #469, opened 2022-07-06
tests: Fix tests on FreeBSD-devel

chrissie-c:
FreeBSD-devel can allocate more space than requested which causes
the check to fail. So check for >= allocated rather than ==


chrissie-c:
10:51 < fabbione> chrissie: win, libqb test passed in 17 minutes
10:51 < fabbione> chrissie: ack from CI overlord

PR #470, opened 2022-07-26
configure: Modernize configure.ac a bit

jfriesse:
... to make 2.71 happy. Also increase minimum version to 2.69 (10 years
old version so should be compatible enough).

There is still one warning left caused by AX_PTHREAD, but this is problem with AX_PTHREAD and fixed upstream (-> it will disappears itself one day)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,7 +1,7 @@
 
 # Process this file with autoconf to produce a configure script.
 
-AC_PREREQ([2.61])
+AC_PREREQ([2.69])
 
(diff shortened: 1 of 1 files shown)


kgaillot:
lgtm

FYI, 2.69 is available in at least RHEL 7, Debian 8 "jessie" (2015), and Ubuntu "xenial" (2016)

chrissie-c:
Looks good to me too. Merged Thanks!

PR #473, opened 2022-08-20
strlcpy: avoid compiler warning from strncpy

wferi:
Otherwise GCC complains about ‘__builtin_strncpy’ specified bound
depends on the length of the source argument.

Signed-off-by: Ferenc Wágner <wferi@debian.org>


chrissie-c (approved):
I'm not sure how this got missed. I need to check my email settings

PR #475, opened 2022-11-03
Add --disable-tests option

ffontaine:
Add `--disable-tests` to allow the user to disable tests. As a side-effect, this will avoid the following build failure when check is found:

```
libstat_wrapper.c:11:10: fatal error: gnu/lib-names.h: No such file or directory
   11 | #include <gnu/lib-names.h>
      |          ^~~~~~~~~~~~~~~~~
```

This build failure is raised since version 2.0.5 and https://github.com/ClusterLabs/libqb/commit/78df90b180740712d0c90b6d982b78241cc99d72

Fixes:
 - http://autobuild.buildroot.org/results/450cfc36d4fd6dc71c138bec45f05b5a2d92a08d

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -39,7 +39,11 @@ ACLOCAL_AMFLAGS		= -I m4
 
 dist_doc_DATA		= COPYING INSTALL README.markdown
 
-SUBDIRS			= include lib doxygen2man docs tools tests examples
+SUBDIRS			= include lib doxygen2man docs tools examples
+
+if ENABLE_TESTS
+SUBDIRS			+= tests
+endif
 
 dist-clean-local:
 	rm -f .snapshot-version autoconf automake autoheader
--- a/configure.ac
+++ b/configure.ac
@@ -562,6 +562,11 @@ AC_ARG_WITH([force-sockets-config-file],
 	[ FORCESOCKETSFILE="$withval" ],
 	[ FORCESOCKETSFILE="$sysconfdir/libqb/force-filesystem-sockets" ])
 
+AC_ARG_ENABLE([tests],
+  [AS_HELP_STRING([--disable-tests],[disable tests])],,
+  [ enable_tests="yes" ])
+AM_CONDITIONAL([ENABLE_TESTS], [test x$enable_tests = xyes])
+
 AC_ARG_ENABLE([install-tests],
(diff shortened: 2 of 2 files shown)


fabbione (approved):

fabbione:
ok to test

fabbione:
> Add `--disable-tests` to allow the user to disable tests. As a side-effect, this will avoid the following build failure when check is found:

I have no objections to add --disable-tests, tho:

> 
> ```
> libstat_wrapper.c:11:10: fatal error: gnu/lib-names.h: No such file or directory
>    11 | #include <gnu/lib-names.h>
>       |          ^~~~~~~~~~~~~~~~~
> ```
> 
> This build failure is raised since version 2.0.5 and [78df90b](https://github.com/ClusterLabs/libqb/commit/78df90b180740712d0c90b6d982b78241cc99d72)

this part I don´t understand. We run CI on each build and we have never seen this problem before.

What distribution / container / platform are you building on? Is the version of check too old?

ffontaine:
The build failure is raised by [buildroot](buildroot.org/) autobuilders.
check is in version 0.15.2 but check doesn't provide `gnu/lib-names.h`.
This file is glibc-specific, it will raise the above build failure with uclibc and musl.

chrissie-c:
I think it would be better just to disable that test. It's a good thing to test libqb against other libcs where possible.

ffontaine:
Indeed, you could disable or remove this test however it still makes sense to disable all tests to save build time and disk space.

fabbione:
> The build failure is raised by [buildroot](buildroot.org/) autobuilders. check is in version 0.15.2 but check doesn't provide `gnu/lib-names.h`. This file is glibc-specific, it will raise the above build failure with uclibc(-ng) and musl.

Apologize for my complete ignorance, but what is buildroot.org?

+1 on disabling the test when gnu/lib-names.h is not available.

Regardless of the libc is in use, test suite should be built and executed. Build only doesn´t help much for the final results. Shaving 3/4 seconds to build is meaningless if the result doesn´t work on the libc in use.

tpetazzoni:
Why is gnu/lib-names.h even included? I don't see anything from this header file that gets used in libstat_wrapper.c. Removing the gnu/lib-names.h solves the build issue.

PR #486, opened 2023-06-01
blackbox: fix potential overlow/memory corruption

chrissie-c:
if the message was too long, then msg_len was added to the buffer size twice, thus causing potential data corruption (seen VERY rarely in the CI test - or, at least, I think it was this).

Also fix a double close() spotted by gcc13's -fanalyzer


jfriesse (approved):
ACK. Nice to see gcc analyzer gets better than coverity ;)

chrissie-c:
retest this please

PR #487, opened 2023-06-06
spec: Migrate to SPDX license

jfriesse:
Both Fedora and openSUSE now recommends to use SPDX shortname format for License.


chrissie-c (approved):

chrissie-c:
Thanks for all your work on these Honza, It's much appreciated!

PR #490, opened 2023-07-19
log: fix potential overflow with long log messages

chrissie-c:
qb_vsnprintf_serialize was called with 'max_size' as the limiting number for the length of the formatted log message. But the buffer also needs to contain the
log header (given by 'actual_size'), so we now pass 'max_size - actual_size' as the maximum length of the formatted log message.

Also added error checks to the blacbkbox calls at
the end of the test, as these now provide a proper test that the BB is functioning. Before they were
masking failures.

### DIFF
--- a/lib/log_blackbox.c
+++ b/lib/log_blackbox.c
@@ -110,8 +110,8 @@ _blackbox_vlogger(int32_t target,
 	chunk += sizeof(uint32_t);
 
 	/* log message */
-	msg_len = qb_vsnprintf_serialize(chunk, max_size, cs->format, ap);
-	if (msg_len >= max_size) {
+	msg_len = qb_vsnprintf_serialize(chunk, t->max_line_length, cs->format, ap);
+	if (msg_len >= t->max_line_length) {
 	    chunk = msg_len_pt + sizeof(uint32_t); /* Reset */
 
 	    /* Leave this at QB_LOG_MAX_LEN so as not to overflow the blackbox */
--- a/tests/check_log.c
+++ b/tests/check_log.c
@@ -832,8 +832,10 @@ START_TEST(test_log_long_msg)
 		qb_log(LOG_INFO, "Message %d %d - %s", lpc, lpc%600, buffer);
 	}
 
-        qb_log_blackbox_write_to_file("blackbox.dump");
-        qb_log_blackbox_print_from_file("blackbox.dump");
+        rc = qb_log_blackbox_write_to_file("blackbox.dump");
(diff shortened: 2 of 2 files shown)


jfriesse (approved):
ACK, really good catch.

jfriesse (approved):
ACK^2 :)

lib/log_blackbox.c
```diff
@@ -110,8 +110,8 @@ _blackbox_vlogger(int32_t target,
 	chunk += sizeof(uint32_t);
 
 	/* log message */
-	msg_len = qb_vsnprintf_serialize(chunk, max_size, cs->format, ap);
-	if (msg_len >= max_size) {
+	msg_len = qb_vsnprintf_serialize(chunk, t->max_line_length, cs->format, ap);
+	if (msg_len >= t->max_line_length) {
 	    chunk = msg_len_pt + sizeof(uint32_t); /* Reset */
 
 	    /* Leave this at QB_LOG_MAX_LEN so as not to overflow the blackbox */
```
kgaillot, line 117: FYI, the next line could theoretically be a problem if the caller used `QB_LOG_CONF_MAX_LINE_LEN` to set the max line length below the length of the "too long" message. (I wouldn't consider it a security issue since the message is hardcoded, not to mention it would be bizarre for a caller to do that). In any case, the "too long" message isn't really helpful -- why not just use as much of the long message as possible (maybe replacing the last three usable characters with an ellipsis)?

jfriesse: +1 for replacing "too long" message with shortened version of logged message. Also I think it may make sense to check QB_LOG_CONF_MAX_LINE_LEN for some minimal reasonable length (right now only maximum length of QB_LOG_ABSOLUTE_MAX_LEN is checked)

chrissie-c:
The ellipsis code has been in there for ages
        rc = qb_log_ctl(t, QB_LOG_CONF_ELLIPSIS, QB_TRUE);

0ec02f9ac589e9e21e447f4406ec104ade01ef73

kgaillot:
> The ellipsis code has been in there for ages, but it needs to be enabled:
> 
> ```
>     rc = qb_log_ctl(t, QB_LOG_CONF_ELLIPSIS, QB_TRUE);
> ```
> 
> [0ec02f9](https://github.com/ClusterLabs/libqb/commit/0ec02f9ac589e9e21e447f4406ec104ade01ef73)

log_blackbox.c doesn't use qb_log_target_format() (I'm not familiar enough with the code to know why not), so the ellipsis code isn't hit

PR #492, opened 2023-11-13
Change assert in ringbuffer to a return code

chrissie-c:
Why just this one?

There are LOADS of asserts in libqb, some are OK and some may be overkill. This one in particular is causing CI failures and so annoys me more than the rest.


jfriesse (approved):
ACK - very specific conditions must be met to make this assert happen, but it is weird when user app asserts unexpectedly.

PR #495, opened 2024-01-29
Make some logs less noisy

kgaillot:
When trying to open an IPC connection, log at info level rather than warning or error, because the caller has better context to know whether it's a problem or not. Users get distracted by error logs that don't actually indicate a problem.

Also lower some spammy messages to trace level.


kgaillot:
Also, I haven't gotten a chance to test this yet, but it's pretty straightforward

PR #496, opened 2024-04-26
Add fdatasync declaration on darwin

viraptor:
Clang 16 starts to explicitly fail on undeclared functions. Unfortunately, while darwin provides the implementation, it's not available in any header.
This can be hacked around by providing a custom signature in the config header. There may be a less terrible way to achieve it.


chrissie-c:
retest this pleae

chrissie-c:
retest this please

chrissie-c (approved):
Looks OK to me, sorry for the delay

PR #506, opened 2025-10-31
Fix a lot of annoying warning compiling with clang.

chrissie-c:
Also remove the really crazy macros used for defining the tests which were hard to read and totally unnecessary


chrissie-c:
retest this please

jfriesse (approved):
ACK. Check macros now finally looks like C. qblist.h - there is probably no other way how to solve the problem without changing API in `queue.h` style.

PR #510, opened 2026-03-17
use #include <poll.h> consistently

statzitz:
Fix some annoying warning on alpine build


chrissie-c (approved):
Seems sensible to me, and passes CI

PR #513, opened 2026-04-24
Silence warnings, ignore *.fdata

andyprice:
Make -Werror more useful during development by cleaning up some warnings. See the commit logs for details.

Also add *.fdata to .gitignore as those files are generated by tests and examples.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,6 @@ test-driver
 /tests/functional/**/log_interlib_client
 /tests/functional/_pkgs
 /tests/functional/_results
+
+# blackbox data dumps from tests/examples
+*.fdata
--- a/configure.ac
+++ b/configure.ac
@@ -223,6 +223,7 @@ AC_TYPE_UINT32_T
 AC_TYPE_UINT16_T
 AC_TYPE_UINT8_T
 
+AC_CHECK_DECLS_ONCE([[__xstat]], [], [], [[#include <sys/stat.h>]])
 AC_CHECK_MEMBER([struct sockaddr_un.sun_len],
                 [AC_DEFINE([HAVE_STRUCT_SOCKADDR_UN_SUN_LEN],
                            1,
--- a/docs/man.dox.in
+++ b/docs/man.dox.in
@@ -1365,14 +1365,9 @@ SKIP_FUNCTION_MACROS   = NO
 # Configuration options related to the dot tool
 #---------------------------------------------------------------------------
 
-# If the CLASS_DIAGRAMS tag is set to YES, doxygen will generate a class diagram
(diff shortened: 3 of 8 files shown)


andyprice:
Ok RHEL 8 CI doesn't like my __xstat fix:
```
CC       _libstat_wrapper.lo
_libstat_wrapper.c:13:12: error: static declaration of ‘__xstat’ follows non-static declaration
static int __xstat(int __ver, const char *__filename, struct stat *__stat_buf)
In file included from _libstat_wrapper.c:8:
 /usr/include/sys/stat.h:397:12: note: previous declaration of ‘__xstat’ was here
extern int __xstat (int __ver, const char *__filename,
```
Rethinking...

chrissie-c:
yeah, xstat is/was a nightmare! Any better ideas for fudging this appreciated :)

chrissie-c:
I like it thank you. Can you also get rid of the doxygen warning too - we might as well silence as much as possible in this patch :)

in docs/man.dox.in it's just a question of replacing
-CLASS_DIAGRAMS         = NO
+CLASS_GRAPH = NO

and maybe the associated comment that goes with it.

andyprice:
> I like it thank you. Can you also get rid of the doxygen warning too - we might as well silence as much as possible in this patch :)
> 
> in docs/man.dox.in it's just a question of replacing -CLASS_DIAGRAMS = NO +CLASS_GRAPH = NO
> 
> and maybe the associated comment that goes with it.

Done. I felt the whole template comment for CLASS_GRAPH was too much to replace the CLASS_DIAGRAMS comment with so I just added a one-line comment instead. I didn't check that RHEL 8 has the CLASS_GRAPH option but I'm sure CI will tell us if it doesn't!

Edit: There were no complaints from RHEL 8

andyprice:
There is still this warning left (which is an info for gcc and a warning for clang...)
```
  CC       print_ver.o
print_ver.c:42:1: warning: "2.0.9.11-84c9" parsed as: "2" "." "0" "." "9" ".11-84c9" [-W#pragma-messages]
   42 | _Pragma(MSG);
      | ^
<scratch space>:254:2: note: expanded from here
  254 |  message ("\"2.0.9.11-84c9\" parsed as: \"2\" \".\" \"0\" \".\" \"9\" \".11-84c9\"")
      |  ^
1 warning generated.

```
I'm not really sure what that's all about but it looks deliberate so I don't intend to touch it in this PR.

chrissie-c:
That warning is there mainly (I think) to prove that the version number has been correctly set in the Makefile. There might be better ways of doing it (or not bothering - nobody else seems to do it. But I agree it's not for this PR

chrissie-c (approved):
ACK thanks!

PR #514, opened 2026-04-30
doxygen2man: Fix segfault on anonymous enums

andyprice:
For these the "name" xml tag has a NULL children field. doxygen2man currently doesn't seem to output anything for enums so we can just check for the NULL pointer and use strdup("enum") as the name in that case.

You can test this easily by patching `enum { FOO };` into `include/qb/qbarray.h` or another header that doxygen2man processes the doxygen xml for, and then running `make`.

I guess this might need some iteration if doxygen2man should be outputting something for enums.

~~There's also a second `strdup(name)` at https://github.com/ClusterLabs/libqb/blob/main/doxygen2man/doxygen2man.c#L1043 which can probably be removed since the name was already `strdup()`ed. But first thing's first :)~~ The latest static analysis in CI failed on that strdup so I've removed it.


chrissie-c (approved):
Makes sense to me. ACK.

PR #515, opened 2026-05-08
configure: add --disable-man option

skirmess:
## Problem

When cross-compiling, `libxml2` is required unconditionally by
`configure`, even though it is only needed by the `doxygen2man` tool
to generate man pages. In a cross-compilation environment (e.g.
OpenWrt), `doxygen2man` cannot be built for the build host and
`libxml2` may not be available as a host library, causing the
configure step to fail.

## Solution

Add a `--disable-man` configure option that skips the `libxml2`
check and the `doxygen2man` build entirely. When disabled,
`BUILD_MAN` is set to false and no man pages are generated.

The default behaviour (`--enable-man`) is unchanged.

## Testing

```
./autogen.sh
./configure --disable-man   # succeeds without libxml2
./configure                 # existing behaviour unchanged
```

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -39,7 +39,11 @@ ACLOCAL_AMFLAGS		= -I m4
 
 dist_doc_DATA		= COPYING INSTALL README.markdown
 
-SUBDIRS			= include lib doxygen2man docs tools examples
+SUBDIRS			= include lib
+if BUILD_MAN
+SUBDIRS			+= doxygen2man
+endif
+SUBDIRS			+= docs tools examples
 
 if ENABLE_TESTS
 SUBDIRS			+= tests
--- a/configure.ac
+++ b/configure.ac
@@ -178,22 +178,30 @@ AC_DEFINE_UNQUOTED([HAVE_GLIB], [1], [We have glib])
 fi
 
 # For building doxygen2man and man pages
-PKG_CHECK_MODULES([libxml], [libxml-2.0])
+AC_ARG_ENABLE([man],
+  [AS_HELP_STRING([--disable-man], [disable man page generation @<:@default=yes@:>@])],
+  [], [enable_man=yes])
 
-# if we are not cross-compiling, we can use the locally built
-# version of doxygen2man, otherwise we can look for
-# a locally installed version. If neither are usable, then
-# don´t build the man pages
-if test "x$cross_compiling" = "xno"; then
-  AM_CONDITIONAL([BUILD_MAN], [true])
-  DOXYGEN2MAN="\$(abs_builddir)/../doxygen2man/doxygen2man"
-else
-  AC_CHECK_PROGS([DOXYGEN2MAN], [doxygen2man])
-  if test "x$DOXYGEN2MAN" = "x"; then
-    AM_CONDITIONAL([BUILD_MAN], [false])
-  else
+if test "x$enable_man" = "xyes"; then
+  PKG_CHECK_MODULES([libxml], [libxml-2.0])
+
+  # if we are not cross-compiling, we can use the locally built
+  # version of doxygen2man, otherwise we can look for
+  # a locally installed version. If neither are usable, then
+  # don't build the man pages
+  if test "x$cross_compiling" = "xno"; then
     AM_CONDITIONAL([BUILD_MAN], [true])
+    DOXYGEN2MAN="\$(abs_builddir)/../doxygen2man/doxygen2man"
+  else
+    AC_CHECK_PROGS([DOXYGEN2MAN], [doxygen2man])
+    if test "x$DOXYGEN2MAN" = "x"; then
+      AM_CONDITIONAL([BUILD_MAN], [false])
+    else
+      AM_CONDITIONAL([BUILD_MAN], [true])
+    fi
   fi
+else
+  AM_CONDITIONAL([BUILD_MAN], [false])
 fi
 AC_SUBST(DOXYGEN2MAN)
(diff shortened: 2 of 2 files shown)


fabbione:
@skirmess the patch itself needs a bit of cleaning, I will leave separate comments for that, but there is a different approach you can use as we did in the past in kronosnet to build doxy2man (before it moved to libqb) in cross-compiler environment:

https://github.com/kronosnet/kronosnet/blob/stable1/configure.ac#L296-L331

this way, you can still generate man pages even if you are not building for native target.

configure.ac
```diff
@@ -862,10 +870,10 @@ AC_CONFIG_FILES([Makefile
 		 tests/Makefile
 		 tests/test.conf
 		 examples/Makefile
-		 doxygen2man/Makefile
 		 docs/Makefile
 		 docs/man.dox
 		 lib/qblog_script.la:lib/qblog_script.la.in])
+AM_COND_IF([BUILD_MAN], [AC_CONFIG_FILES([doxygen2man/Makefile])])
```
fabbione: this change is unnecessary. even if cross-compiling you still want to have the full ability to run make dist and make clean. Skipping on Makefile will leave doxygen2man out of a tarball or dirty.

Makefile.am
```diff
@@ -39,7 +39,11 @@ ACLOCAL_AMFLAGS		= -I m4
 
 dist_doc_DATA		= COPYING INSTALL README.markdown
 
-SUBDIRS			= include lib doxygen2man docs tools examples
+SUBDIRS			= include lib
```
fabbione: overly complex. just keep include lib docs tools examples in one line and add doxygen2man conditionally, like it´s done for tests just few lines below.

skirmess:
Thanks for the pointer to the kronosnet approach. I've addressed the inline review comments in the updated commit.                                                                              
                                                                                                                                                                                                                     
Regarding the kronosnet approach: it requires libqb to be installed natively on the build host (the PKG_CHECK_MODULES([libqb_BUILD], [libqb]) call uses the host's pkg-config). In the OpenWrt SDK environment - which is the use case motivating this PR - libqb is only cross-compiled for the target and is never installed on the build host. So that approach unfortunately doesn't apply here.                                
                                                                                                                                                                                                                     
Additionally, OpenWrt does not ship a man command, so man pages on the target would not be useful regardless.                                                                                                      
                                                                                                                                                                                                                   
`--disable-man` is the pragmatic solution for cross-compilation environments where the build host does not have libqb natively available. The kronosnet approach could be a useful follow-up enhancement for environments where a native libqb is present, but it would not help the OpenWrt case.

fabbione:
@skirmess thanks for addressing the changes. I am ok with this approach.

fabbione (approved):

chrissie-c:
This is failing all over our CI, please fix https://ci.kronosnet.org/job/libqb/job/libqb-pipeline//job/PR-515/2/artifact/FAILED_voting-fedora-43-x86-64.log

skirmess:
I've pushed an updated commit addressing the inline feedback.

One note on the Makefile.am ordering: adding `doxygen2man` at the end with `+=` causes a build failure — `docs` is processed first and tries to invoke `doxygen2man/doxygen2man` before it has been built:

```
/bin/sh: line 1: .../doxygen2man/doxygen2man: No such file or directory
make[2]: *** [Makefile:931: xml-man] Error 127
```

Unlike tests, `doxygen2man` must be built before `docs` since `docs` invokes it directly. The ordering therefore needs to insert `doxygen2man` between `lib` and `docs` when `BUILD_MAN` is true:

```makefile
SUBDIRS = include lib
if BUILD_MAN
SUBDIRS += doxygen2man
endif
SUBDIRS += docs tools examples
```

`doxygen2man/Makefile` is still generated unconditionally so that `make dist` and `make distclean` work correctly in all cases.

fabbione:
> I've pushed an updated commit addressing the inline feedback.
> 
> One note on the Makefile.am ordering: adding `doxygen2man` at the end with `+=` causes a build failure — `docs` is processed first and tries to invoke `doxygen2man/doxygen2man` before it has been built:
> 
> ```
> /bin/sh: line 1: .../doxygen2man/doxygen2man: No such file or directory
> make[2]: *** [Makefile:931: xml-man] Error 127
> ```
> 
> Unlike tests, `doxygen2man` must be built before `docs` since `docs` invokes it directly. The ordering therefore needs to insert `doxygen2man` between `lib` and `docs` when `BUILD_MAN` is true:

if doc requires doxygen2man, then --disable-man will fail because docs are built unconditionally? I didn´t test it but I assume you did?

> 
> ```makefile
> SUBDIRS = include lib
> if BUILD_MAN
> SUBDIRS += doxygen2man
> endif
> SUBDIRS += docs tools examples
> ```
> 
> `doxygen2man/Makefile` is still generated unconditionally so that `make dist` and `make distclean` work correctly in all cases.

skirmess:
Good question — I tested `./configure` but not `make`. I've now run both:

`docs/Makefile.am` guards all doxygen2man invocations behind `if BUILD_MAN`, so when `--disable-man` is set, make descends into `docs` but finds nothing to do:

```
Making all in docs
make[2]: Nothing to be done for 'all'.
```

Full build with `--disable-man` completes successfully without requiring doxygen2man or libxml2.

chrissie-c:
Committed. thanks.

skirmess:
Thank you @fabbione and @chrissie-c .

PR #516, opened 2026-05-12
strchr() const correctness fixes

andyprice:
Clang now warns when the variable storing the result of strchr() throws away the constness of the string passed into it.

For remove_tempdir() and qb_sys_mmap_file_open() the fix is a simple addition of const.

qb_log_thread_log_write() needs const removed from its buffer parameter as the qb_do_extended macro writes to it when handling the special extended logging marker. Luckily nothing passes a const buffer to the function so that's okay.


chrissie-c (approved):
Looks good to me, thanks!

PR #517, opened 2026-05-21
Misc claude fixes

chrissie-c:
A miscellany of small fixes found by everyone's enemy

### DIFF
--- a/lib/ipc_setup.c
+++ b/lib/ipc_setup.c
@@ -371,6 +371,14 @@ qb_ipc_auth_creds(struct ipc_auth_data *data)
 			if (cmsg->cmsg_type != SCM_CREDENTIALS)
 				continue;
 
+			/* Validate that the control message contains a full ucred structure */
+			if (cmsg->cmsg_len < CMSG_LEN(sizeof(struct ucred))) {
+				qb_util_log(LOG_WARNING,
+					    "received malformed credential message (len=%zu, expected=%zu)",
+					    (size_t)cmsg->cmsg_len, CMSG_LEN(sizeof(struct ucred)));
+				continue;
+			}
+
 			memcpy(&cred, CMSG_DATA(cmsg), sizeof(struct ucred));
 			res = 0;
 			data->ugp.pid = cred.pid;
--- a/lib/ipc_setup.c
+++ b/lib/ipc_setup.c
@@ -716,8 +716,11 @@ handle_new_connection(struct qb_ipcs_service *s,
 		res = -errno;
 		goto send_response;
 	}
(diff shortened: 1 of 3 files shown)


jfriesse (approved):
ACK. Honestly, I still cannot believe this is written by claude explode. It's small, it's targeted, there are no emojis, long checkmark lists, no non-sense comments. And all good catches.

lib/ipc_socket.c
```diff
@@ -816,17 +816,20 @@ qb_ipcs_us_connect(struct qb_ipcs_service *s,
...
 	(void)strlcpy(c->request.u.us.shared_file_name, r->request, NAME_MAX);
+	/* These aren't failures, but are worth logging */
 	res = chown(r->request, c->auth.uid, c->auth.gid);
 	if (res != 0) {
-		/* ignore res, this is just for the compiler warnings.
-		 */
 		res = 0;
+		qb_util_perror(LOG_ERR, "failed to chown shared memory file (%s)",
+			       r->request);
+		goto cleanup_hdr;
```
wferi, line 825: On one hand, these aren't strdup() checks, and on the other, they change behavior by not ignoring the chmod()/chown() failures anymore, contrary to the added comment. (Sorry if I'm talking nonsense, just pattern-matching here without reasonable familiarity with the code, but I got intrigued by Jan's comment.)

jfriesse: @wferi Good catch. Yes, I was overoptimistic and haven't noticed it changes behavior :( There should be really either change of comment or (probably better) removal of `goto cleanup_hdr`;

lib/ipc_setup.c
```diff
@@ -708,8 +716,11 @@ handle_new_connection(struct qb_ipcs_service *s,
 		res = -errno;
 		goto send_response;
 	}
-	/* chown can fail because we might not be root */
-	(void)chown(c->description, c->auth.uid, c->auth.gid);
+	/* chown may fail if not root, but log it */
+	if (chown(c->description, c->auth.uid, c->auth.gid) != 0) {
+		qb_util_perror(LOG_WARNING, "failed to chown directory (%s)",
+			       c->description);
```
gao-yan, line 722: Hi, with this, the pacemaker sub-daemons running as `haclient` keep reporting this warning, even though the IPC connections seem to be working. 

The description of the commit https://github.com/ClusterLabs/libqb/pull/381/changes/700fb2b27eccf16c61bfee26aa555d4e77267306 also says "don't abort if we aren't permitted to chown() it.  The client might still have the privileges to enter it."

Maybe we don't need to warn about this, or at least we could lower its logging priority?

PR #520, opened 2026-06-01
Fix building using the LCC compiler

arrowd:
Just like Clang, LCC (a closed source compiler for the Elbrus2000 arch) does not support VLAs in structure definitions.


chrissie-c:
retest this please

chrissie-c (approved):
There's an argument that this ifdef should now be an ifdef gcc as it is the odd one out now. But as lcc is a closed source compiler I'm happy to leave it as a 'weirdo' :)

ACK Thanks.

